PR #6, opened 2018-08-07
[IMP] account_financial_risk: Only block positive customer invoices validation

carlosdauden:
@Tecnativa

### DIFF
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -11,7 +11,8 @@ class AccountInvoice(models.Model):
     def action_invoice_open(self):
         if self.env.context.get('bypass_risk', False):
             return super(AccountInvoice, self).action_invoice_open()
-        for invoice in self:
+        for invoice in self.filtered(lambda x: x.type in (
+                'out_invoice', 'out_refund') and x.amount_total_signed > 0.0):
             partner = invoice.partner_id.commercial_partner_id
             exception_msg = ""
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You should not limit invoices that reduces the risk. They are:

- Normal invoices with negative amount.
- Refund invoices with positive amount.

carlosdauden:
Those invoices that you comment, in addition to inconsistent, can cause other more serious problems that don't show the risk exceeded warning :wink:

pedrobaeza:
That invoices can exist, so we should cover them. Don't think only on Spanish regulation (that are also admitted in some cases).

carlosdauden:
Changes done

account_financial_risk/models/account_invoice.py
```diff
@@ -11,7 +11,8 @@ class AccountInvoice(models.Model):
     def action_invoice_open(self):
         if self.env.context.get('bypass_risk', False):
             return super(AccountInvoice, self).action_invoice_open()
-        for invoice in self:
+        for invoice in self.filtered(lambda x: x.type in (
+                'out_invoice', 'out_refund') and x.amount_total_signed > 0.0):
```
pedrobaeza, line 15: This should be `<=`

pedrobaeza: Forget it. I have mixed the condition.

pedrobaeza (approved):
Fast-tracking this.

PR #8, opened 2018-09-11
[11.0][IMP] account_financial_risk: Add policy states

sergio-teruel:
cc @Tecnativa


pedrobaeza:
Why this in this module? This should be in the other one `partner_risk_insurance`.

PR #12, opened 2019-02-22
12.0 mig account financial risk

ernestotejeda:
Cc @Tecnativa


account_financial_risk/templates/assets.xml
```diff
@@ -3,7 +3,7 @@
 
     <template id="assets_common" inherit_id="web.assets_common">
         <xpath expr="." position="inside">
-            <link rel="stylesheet" type="text/css" href="/account_financial_risk/static/src/less/account_financial_risk_table.less"/>
+            <link rel="stylesheet" type="text/css" href="/account_financial_risk/static/src/scss/account_financial_risk_table.scss"/>
```
pedrobaeza: You must convert also the syntax. It's not the same for LESS than for SCSS.

account_financial_risk/models/res_company.py
```diff
@@ -9,7 +9,7 @@ class ResCompany(models.Model):
 
     invoice_unpaid_margin = fields.Integer(
         string="Maturity Margin",
-        help="Days after due date to set an invoice as unpaid."
-             "The change of this field recompute all partners risk,"
+        help="Days after due date to set an invoice as unpaid. "
```
pedrobaeza, line 12: If you change this text, then you should update the msgid in all the PO/POT files.

pedrobaeza (changes requested):
This is a special case of a module being renamed in the middle of its story, so half of the commits are not here. It's not your fault and the migration guide doesn't cover this eventual situation, but we can bring all the history and amend it for v13 and so on to be fetched with the migration guide.

Let me put you the commands from the beginning to do it for being complete, assuming you have your SSH keys declared in GitHub, but for sure next time you have to do it you will know :) :
```
git clone git@github.com:OCA/credit-control -b 12.0
cd credit-control
git remote add tecnativa git@github.com:Tecnativa/credit-control.git
git fetch tecnativa
git checkout -b 12.0-mig-account_financial_risk
# old-name
git format-patch --keep-subject --stdout origin/12.0..df8c52c29da4f5211fb752823e3c33cd3e41bfc8 -- partner_financial_risk | git am -3 --keep
# new name
git format-patch --keep-subject --stdout origin/12.0..origin/11.0 -- account_financial_risk | git am -3 --keep
# Rewrite history for having all with the new name for future migrations
git filter-branch --tree-filter 'if [ -d partner_financial_risk ]; then mv partner_financial_risk account_financial_risk; fi' HEAD
# Your migration commit
git cherry-pick e1d314c2221c80ebe17c7cfc9157c7aaa496517c
git push --set-upstream tecnativa -f
```
Can you try and let me know?

pedrobaeza (approved):

yajo (approved):
Code OK

PR #14, opened 2019-03-07
12.0 mig sale_financial_risk

ernestotejeda:
Cc @Tecnativa


pedrobaeza (commented):
The new field is very heavy to be compute. Can we light this up a bit with some `read_group` operations?

sale_financial_risk/models/sale.py
```diff
@@ -32,3 +32,79 @@ def action_confirm(self):
...
+
+            Then, compute amount to invoice as:
+                total_sale_line - amount_invoiced
+            where
+                `amount_invoiced` is taxed amount previously explained
+        """
+        refund_lines_product = self.env['account.invoice.line']
+        for line in self:
+            invoice_lines = line.invoice_lines.filtered(
+                lambda l: l.invoice_id.state in ['open', 'in_payment', 'paid'])
```
yajo: Use `{'open', 'in_payment', 'paid'}` for performance.

sale_financial_risk/models/sale.py
```diff
@@ -32,3 +32,79 @@ def action_confirm(self):
...
+            # Refund invoices linked to invoice_lines
+            refund_invoices = invoice_lines.mapped(
+                'invoice_id.refund_invoice_ids').filtered(
+                lambda inv: inv.state in ('open', 'in_payment', 'paid'))
```
yajo: Use `{'open', 'in_payment', 'paid'}` for performance.

sale_financial_risk/models/sale.py
```diff
@@ -32,3 +32,79 @@ def action_confirm(self):
...
+            for ref_line in refund_invoice_lines:
+                ref_date = (ref_line.invoice_id.date_invoice
+                            or fields.Date.today())
+                if ref_line.invoice_id.type == 'out_refund':
+                    amount_invoiced -= ref_line.currency_id._convert(
+                        ref_line.price_total, line.currency_id,
+                        line.company_id, ref_date)
+
+            amount_to_invoice = 0.0
+            if line.state in ['sale', 'done']:
```
yajo: Use `{'sale', 'done'}` for performance.

yajo (approved):
The new field seems expensive, but at least it's stored, so it shouldn't be much of a problem.

The `read_group` isn't a panacea: it saves some Python loops, but at the cost of increasing SQL queries (because you can't benefit of caching and prefetching). I think we can defer that "fix" until we see it's actually a problem.

Code OK, just some little obvious remarks for a little better performance.

sale_financial_risk/models/sale.py
ernestotejeda: Done

sale_financial_risk/models/sale.py
ernestotejeda: Done

sale_financial_risk/models/sale.py
ernestotejeda: Done, thank you @Yajo

pedrobaeza:
@carlosdauden your last call about the method's performance.

@ernestotejeda please cover with tests all the code for not dropping coverage.

ernestotejeda:
> @ernestotejeda please cover with tests all the code for not dropping coverage.

Done

carlosdauden (approved):

PR #16, opened 2019-03-14
[12.0][MIG] account_credit_control

MiquelRForgeFlow:
Moved from https://github.com/OCA/account-financial-tools/pull/779.


pedrobaeza:
@mreficent does #18 's comments applied here?

account_credit_control/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Account Credit Control module is a part of Financial Tools used in business to
+ensure that once sales are made they are realised as cash. This module helps to
+identify outstanding debt beyond tolerance level and setup followup method.
+
+Configuration
+=============
+
+Configure the policies and policy levels in ``Accounting  > Configuration >
+Credit Control > Credit Control Policies``.
```
cubells, line 17: Menú is now: ``Invoicing > Configuration > Credit Control > Credit Control Policies``

account_credit_control/README.rst
```diff
@@ -0,0 +1,86 @@
...
+You can define as many policy levels as you need.
+
+Configure a tolerance for the Credit control and a default policy
+applied on all partners in each company, under the Accounting tab.
+
+You are able to specify a particular policy for one partner or one invoice.
```
cubells, line 25: This sentence is not true now. You can't change credit control policy in invoice.

And I think this feature is necessary: to be able to change the credit control policy for a particular invoice.
Or at least to see which credit control policy is applied to any invoice.

account_credit_control/models/credit_control_policy.py
```diff
@@ -0,0 +1,389 @@
...
+         ('previous_date', 'Previous Reminder')],
+        string='Compute Mode',
+        required=True
+    )
+    delay_days = fields.Integer(string='Delay (in days)', required=True)
+    email_template_id = fields.Many2one('mail.template',
+                                        string='Email Template',
+                                        required=True)
+    channel = fields.Selection([('letter', 'Letter'),
+                                ('email', 'Email')],
```
cubells: Selection items are repeated twice, in attribute `channel` in model `credit.control.line` and here in model `credit.control.policy.level` model.

I think we must create a list and reference it here to avoid to have to change this values twice if we increase values in future.

cubells (changes requested):

account_credit_control/README.rst
MiquelRForgeFlow: I just copied the readme of your PR. Is it ok now?

MiquelRForgeFlow:
@pedrobaeza I think it's ready!

pedrobaeza:
Well, everything is red...

MiquelRForgeFlow:
Yeah, now is green 😁

account_credit_control/models/credit_control_run.py
```diff
@@ -209,7 +210,7 @@ def run_channel_action(self):
             comms = comm_obj._generate_comm_from_credit_lines(email_lines)
             comms._generate_emails()
         if letter_lines:
-            wiz = self.env['credit.control.printer'].create({
+            wiz = self.env['credit.control.printer'].create([{
```
pedrobaeza: This doesn't benefit from batch create, so no need to change this.

account_credit_control/tests/test_account_invoice.py
```diff
@@ -6,47 +6,46 @@
...
         Test the method action_cancel on invoice
         We will create an old invoice, generate a control run
         and check if I can unlink this invoice
-        :return:
         """
         journal = self.env['account.invoice']._default_journal()
 
         account_type_rec = self.env.ref('account.data_account_type_receivable')
-        account = self.env['account.account'].create({
+        account = self.env['account.account'].create([{
```
pedrobaeza: Why adding batch create?

pedrobaeza (commented):
Why have you changed all `create` calls with lists? There's no performance benefit for one record, and makes the diff very big.

account_credit_control/tests/test_account_invoice.py
MiquelRForgeFlow: because `account.account` uses the batch create

account_credit_control/models/credit_control_run.py
MiquelRForgeFlow: I have not changed all creates. Just the ones that their model have a _default_ batch create

account_credit_control/tests/test_account_invoice.py
pedrobaeza: But the use is transparent. You don't need to call it in a list AFAIK

MiquelRForgeFlow: If you don't call it in a list, then the api of odoo will put it in a list before doing the create. It's the same.

pedrobaeza: Yes, that's why I ask you to not add ~40 lines of diff because of this, as it adds no value.

MiquelRForgeFlow: The lines affected because of this are fewer than 40 if I remember correctly. It's surprising me that you are trying to justify your point just by mentioning this few lines diff when this PR has a LOT of lines diff due to other reasons, the most part due to putting each attribute of a declared field in a different line.

pedrobaeza: My justification is the same always: don't touch what is not needed to be touched unless there's a good reason for doing it, and this one is not.

MiquelRForgeFlow: The change of putting each attribute of a declared field in a different line it's not needed, but it's done for readability reasons. This case is the same, with a list, you as a reader, understands that `account.account` model has by default the creation batch and if in the future a new tests is added, then the people adding it may have that in mind. If not, then the people will always use the create without the batch.
But for example, the `account.invoice` model doesn't have the creation batch by default. Thus, in the cases of creating an `account.invoice` in this module, you can see that they are created with a dict, without applying the batch.

pedrobaeza: No, this change is not for readability. It's totally unneeded. There's no difference between passing a list or only one value. The implementation might change and won't affect this.  Other thing is if you create several accounts, then you can benefit from do all in batch.

MiquelRForgeFlow: Reverted the changes as you mentioned. You should be happy I always end doing what you want xD

pedrobaeza: Yes, very happy :smiley: . Your contributions are of course valuable, and I always appreciate a constructive dialog.

pedrobaeza: I count 132 lines of diff in the last forced push

More than 40! :stuck_out_tongue_winking_eye:

MiquelRForgeFlow: You are duplicating and thus counting wrong. 68 additions and 68 deletions. It means 68 lines changed!

BTW, when I said less that 40, I was referring only to the `account.account` ones.

pedrobaeza: No, that's an important thing! In fact, you have 132 lines of diff, 66 removing a line, and 66 adding another one "slightly different", but this something that git doesn't matter. In terms of commit size, it counts double. Each change means 2 diff lines.

MiquelRForgeFlow: good to know

account_credit_control/README.rst
pedrobaeza: So the problem seems to be that the feature has been lost. Can you check it, @mreficent ?

pedrobaeza (approved):
Pending on @cubells' comment.

account_credit_control/README.rst
MiquelRForgeFlow: I just checked. In v11 you also can't specify a particular policy for one invoice. It's because in the `account.invoice` form you have `<field name="credit_policy_id" attrs="{'invisible': [('credit_policy_id', '=', False)]}"/>`  😂 So you can never fill this field if the field is invisible, because by default it's always False.

MiquelRForgeFlow: @pedrobaeza @cubells any recommendation in how to amend this? I mean, I could put a non False default, I could delete or change the invisible domain, or maybe other things. What you propose?

pedrobaeza: I would remove that `attrs`. That way, you can change the policy without problems. I don't see the use case for allowing to change only in case there's other previously set.

account_credit_control/views/credit_control_line.xml
```diff
@@ -0,0 +1,191 @@
...
+                <field name="balance_due"/>
+                <field name="currency_id"/>
+                <field name="move_line_id"/>
+                <field name="account_id"/>
+                <field name="policy_level_id"/>
+                <field name="policy_id"/>
+                <field name="mail_message_id"/>
+                <field name="activity_date_deadline"/>
+                <button name="button_schedule_activity" type="object"
+                        aria-label="Schedule activity" title="Schedule activity" icon="fa fa-clock-o"/>
```
astirpe:
```suggestion
                        aria-label="Schedule activity" title="Schedule activity" icon="fa-clock-o"/>
```

MiquelRForgeFlow: ??

astirpe: Otherwise the icon is not displayed.

pedrobaeza:
Merging as comment resolved.

PR #17, opened 2019-03-25
[IMP] account_financial_risk: Manage risk in multiple invoice confirmation wizard

carlosdauden:
@Tecnativa

### DIFF
--- a/account_financial_risk/i18n/es.po
+++ b/account_financial_risk/i18n/es.po
@@ -9,11 +9,10 @@ msgid ""
 msgstr ""
 "Project-Id-Version: partner-contact (9.0)\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-10-08 23:36+0200\n"
-"PO-Revision-Date: 2017-10-08 23:39+0200\n"
+"POT-Creation-Date: 2019-03-25 14:55+0100\n"
+"PO-Revision-Date: 2019-03-25 15:00+0100\n"
 "Last-Translator: Carlos Dauden <carlos.dauden@tecnativa.com>\n"
-"Language-Team: Spanish (http://www.transifex.com/oca/OCA-partner-contact-9-0/"
-"language/es/)\n"
+"Language-Team: Spanish (http://www.transifex.com/oca/OCA-partner-contact-9-0/language/es/)\n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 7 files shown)


account_financial_risk/i18n/es.po
```diff
@@ -423,3 +424,29 @@ msgstr "account.config.settings"
 #: model:ir.model,name:account_financial_risk.model_partner_risk_exceeded_wiz
 msgid "partner.risk.exceeded.wiz"
 msgstr "partner.risk.exceeded.wiz"
+
+#. module: account_financial_risk
+#: model:ir.model,name:account_financial_risk.model_res_config_settings
+msgid "res.config.settings"
+msgstr ""
+
+#~ msgid "Continue method"
```
pedrobaeza: Please don't use POEdit or configure it properly for not letting this residual lines (and check also line wrapping).

account_financial_risk/wizards/account_invoice_state_view.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="account_invoice_confirm_view" model="ir.ui.view">
+        <field name="name">account.invoice.confirm.form</field>
+        <field name="model">account.invoice.confirm</field>
+        <field name="inherit_id" ref="account.account_invoice_confirm_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//footer" position="before">
+                <p attrs="{'invisible': [('info_risk','=', '')]}">
```
pedrobaeza, line 10: Does this work with empty fields (instead of `('info_risk', '=', False)`)?

account_financial_risk/models/account_invoice.py
```diff
@@ -7,32 +7,37 @@
...
-            # If risk_invoice_draft_include this invoice included in risk_total
-            elif not partner.risk_invoice_draft_include and (
-                    partner.risk_invoice_open_include and
-                    (partner.risk_total + invoice.amount_total) >
-                    partner.credit_limit):
-                exception_msg = _(
-                    "This invoice exceeds the financial risk.\n")
+            exception_msg = invoice.risk_exception_msg()
             if exception_msg:
                 return self.env['partner.risk.exceeded.wiz'].create({
```
pedrobaeza, line 38: If we do `return`, this will abort massive confirmation batch.

pedrobaeza (changes requested):

account_financial_risk/wizards/account_invoice_state_view.xml
carlosdauden: https://github.com/OCA/credit-control/pull/17/files#diff-9a2973bb509867614553981c26e976ecR15

account_financial_risk/models/account_invoice.py
carlosdauden: This return isn't modify in this PR

pedrobaeza: Yes, but this is not the expected behavior, isn't it? Or batch confirm call other method?

carlosdauden: This is called one to one in `invoice_confirm()`:
https://github.com/odoo/odoo/blob/9094dd0c8279a8eef582cd5daedefbcfc4af8a21/addons/account/wizard/account_invoice_state.py#L22

account_financial_risk/i18n/es.po
carlosdauden: Changes done

account_financial_risk/wizards/account_invoice_state_view.xml
carlosdauden: `('info_risk', '=', False)` don't work

account_financial_risk/models/account_invoice.py
pedrobaeza: OK

pedrobaeza (approved):

martintorresvm (approved):

pedrobaeza:
Cherry-picked for v12 in 2125302af1ea20b05dcbb99390524273644b5948

PR #22, opened 2019-07-17
[12.0] [ADD] Auto process lower levels

qgroulard:
Add the possibility to mark a policy so that processing a line generated by this policy
will also process lower level lines for the same partner (and policy).

Also add a filter for credit control line tree view in order to only display
highest level lines for each partner and policy.

This helps to manage high credit control line generation rates (e.g. if you are selling
monthly subscriptions).
In such situation you may want to avoid spamming reminders to customers with
several credit control lines of different levels, and only take action for the
highest level. Thus, an email sent for a high level will also contain the lower
level lines.
Plus, you can use the filter ``Group Lines`` in the ``Credit Control Lines``
menu to hide lines that will be auto-processed when a highest level line is.


pedrobaeza:
It would be interesting that you explain in the README the use case for this new option.

sbidoul:
cc/ @fclementic2c

account_credit_control/models/credit_control_line.py
```diff
@@ -247,6 +259,15 @@ def unlink(self):
                     _('You are not allowed to delete a credit control '
                       'line that is not in draft state.')
                 )
+            if line.policy_id.auto_process_lower_levels:
+                related_lines = line.get_related_lines().filtered(
+                    lambda l: l.id != line.id
+                )
+                if not related_lines:
```
rousseldenis: @qgroulard Maybe a helper function to reuse code that writes on related_lines ?

account_credit_control/models/credit_control_policy.py
```diff
@@ -44,6 +53,46 @@ class CreditControlPolicy(models.Model):
...
+    def _compute_yield_by_level(self):
+        for rec in self:
+            rec.yield_by_level = not rec.auto_process_lower_levels
+
+    @api.multi
+    def write(self, values):
+        res = super(CreditControlPolicy, self).write(values)
+        if 'auto_process_lower_levels' in values:
+            credit_control_model = self.env['credit.control.line']
+            if values['auto_process_lower_levels']:
```
rousseldenis: @qgroulard Maybe juste one search would do the job (as the domain is the same):

```suggestion
            # As you don't need fields (just recordset with ids)
            new_self = self.with_context(prefetch_fields=False)
            lines = new_self.search(<with a function to retrieve domain?>)
            if values['auto_process_lower_levels']:
```

rousseldenis (changes requested):
Code review. Just a little change.

account_credit_control/models/credit_control_line.py
```diff
@@ -151,6 +153,16 @@ class CreditControlLine(models.Model):
...
         store=True,
     )
+    auto_process = fields.Selection(
+        selection=[
+            ('no_auto_process', 'No Auto Process'),
+            ('low_level', 'Low Level'),
+            ('highest_level', 'Highest Level'),
+        ],
+        default="no_auto_process",
+        store=True,
```
florian-dacosta: Is it really necessary to specify the field is stored since it is not a computed/related field?

account_credit_control/models/credit_control_line.py
```diff
@@ -256,15 +277,68 @@ def write(self, values):
             self.partner_id.write({
                 'manual_followup': values.get('manual_followup'),
             })
+        for line in self:
+            if (
```
florian-dacosta: I wonder about performance here. If we have 10 related lines and we write something on them all, on each loop we are going to do the same 2 writes.
I think it would be nice to have a system to avoid this behavior. Maybe save a key like (partner_id, currency_id, policy_id) in a list and skip the whole logic if key is already processed?
This is valid for this method but also unlink and create, in my opinion.

account_credit_control/models/credit_control_line.py
```diff
@@ -256,15 +277,68 @@ def write(self, values):
...
+
+    def get_lower_related_lines(self):
+        """
+        Return lines that will receive the same treatment
+        (ie lines of lower level from the same group if grouped).
+        """
+        self.ensure_one()
+        if self.policy_id.auto_process_lower_levels:
+            return self.get_related_lines().filtered(
+                lambda l: l.policy_level_id.level <= self.policy_level_id.level
```
florian-dacosta: You can use directly the level field since it is a related on policy_level_id.level

account_credit_control/models/credit_control_policy.py
```diff
@@ -28,6 +28,15 @@ class CreditControlPolicy(models.Model):
...
              'generate lines or are obsolete',
     )
+    auto_process_lower_levels = fields.Boolean(
+        string='Auto Process Lower levels',
+        help='When an action is performed on a credit control line generated '
+        'by this policy, lower level lines for the same partner and policy '
+        'will also be processed.',
+    )
+    yield_by_level = fields.Boolean(
+        compute='_compute_yield_by_level'
```
florian-dacosta: Is this field really useful? I am not sure to understand why since it is just the oposite of `auto_process_lower_levels` (so I would not compute a field just for this) and I don't see it used in the code.

account_credit_control/models/credit_control_policy.py
```diff
@@ -44,6 +53,46 @@ class CreditControlPolicy(models.Model):
...
+                    domain=[
+                        ('policy_id', '=', self.id),
+                        ('state', 'in', ('draft', 'to_be_sent')),
+                    ],
+                    fields=['id'],
+                    groupby=['partner_id'],
+                )
+                for group in res:
+                    records = credit_control_model.search(
+                        group['__domain'],
```
florian-dacosta: I don't see any sign of `currency_id` in the domain, but it is present `get_related_lines`.
From what I understand it is not consistent.
Since you can't really put the domain of this method and the one of `get_related_lines` in same method you would use in both case, it would be nice to have an automatic test on these 2 methods to be sure it stay consistent even if someone make a change (and could forget easily to put it in both methods)

account_credit_control/models/credit_control_line.py
florian-dacosta: I also think it would be nice to have a helper here since there is some code reused several times.

florian-dacosta (changes requested):
Code review only, I had not time to test it yet.
A last remark, if feasible easily, I think it would be better in a sub module.
Indeed, I don't think it will be a feature used in the majority of cases and the base module is already kind of big, so for a better understanding of the module and to ease migration, it could be nice.
I would not block the PR for this though.

account_credit_control/models/credit_control_line.py
```diff
@@ -247,6 +259,15 @@ def unlink(self):
                     _('You are not allowed to delete a credit control '
                       'line that is not in draft state.')
                 )
+            if line.policy_id.auto_process_lower_levels:
+                related_lines = line.get_related_lines().filtered(
```
sbejaoui: it will be more performant if you add an argument to exclude self from the get_related_lines search

account_credit_control/models/credit_control_line.py
```diff
@@ -247,6 +259,15 @@ def unlink(self):
...
+                    lambda l: l.id != line.id
+                )
+                if not related_lines:
+                    continue
+                related_lines[0].write({'auto_process': 'highest_level'})
```
sbejaoui: What about a method `get_highest_level_from_related_lines`?

account_credit_control/models/credit_control_line.py
```diff
@@ -247,6 +259,15 @@ def unlink(self):
...
+                if len(related_lines) > 1:
+                    related_lines[1:].write({'auto_process': 'low_level'})
```
sbejaoui: `get_related_lines(exclude=line.get_highest_level_from_related_lines())` ?

account_credit_control/models/credit_control_line.py
```diff
@@ -256,15 +277,68 @@ def write(self, values):
...
+            return self
+
+    def get_lower_related_lines(self):
+        """
+        Return lines that will receive the same treatment
+        (ie lines of lower level from the same group if grouped).
+        """
+        self.ensure_one()
+        if self.policy_id.auto_process_lower_levels:
+            return self.get_related_lines().filtered(
```
sbejaoui: a search will be better the a filter.

account_credit_control/models/credit_control_line.py
```diff
@@ -292,3 +366,14 @@ def button_credit_control_line_form(self):
         action['views'] = [(form.id, 'form')]
         action['res_id'] = self.id
         return action
+
+    def act_show_auto_process_line(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "name": "Credit Control Lines",
```
sbejaoui:
```suggestion
            "name": _("Credit Control Lines"),
```

account_credit_control/models/credit_control_line.py
florian-dacosta: You could just do something like 
`related_lines = line.get_related_lines() - line`
no ?

account_credit_control/models/credit_control_line.py
```diff
@@ -159,6 +161,15 @@ class CreditControlLine(models.Model):
...
         store=True,
     )
+    auto_process = fields.Selection(
+        selection=[
+            ('no_auto_process', 'No Auto Process'),
+            ('low_level', 'Low Level'),
+            ('highest_level', 'Highest Level'),
+        ],
+        default="no_auto_process",
+        readonly=True,
```
sbidoul, line 178: A help explaining what the 3 values mean would be helpful, if only as documentation of the algorithm.

account_credit_control/models/credit_control_line.py
```diff
@@ -251,6 +262,18 @@ def create_or_update_from_mv_lines(self, lines, level, controlling_date,
 
         return new_lines
 
+    def update_auto_process(self, exclude_ids=None):
```
sbidoul: Does this method need to be rpc-accessible? If not, prefix it with underscore.

account_credit_control/models/credit_control_line.py
```diff
@@ -268,8 +292,64 @@ def write(self, values):
...
+    def create(self, vals_list):
+        lines = super(CreditControlLine, self).create(vals_list)
+        for line in lines:
+            if line.state == 'sent':
+                line.write({'auto_process': 'no_auto_process'})
+            else:
+                line.update_auto_process()
+        return lines
+
+    def get_highest_related_line(self, exclude_ids=None):
```
sbidoul:
```suggestion
    def _get_highest_related_line(self, exclude_ids=None):
```

account_credit_control/models/credit_control_line.py
```diff
@@ -268,8 +292,64 @@ def write(self, values):
...
+        self.ensure_one()
+        return self.get_related_lines(exclude_ids=exclude_ids, limit=1)
+
+    def get_related_lines(self, exclude_ids=None, limit=None):
```
sbidoul:
```suggestion
    def _get_related_lines(self, exclude_ids=None, limit=None):
```

account_credit_control/models/credit_control_line.py
sbidoul: yes, since `get_related_lines` already does a search it will be more efficient to add a level criteria in the search. OTOH it's probably marginal since the number of related lines will never be huge.

sbidoul (commented):
A few minor comments. Could you also reply to other reviewer comments? It looks like most of them have been handled but I'm not too sure.

account_credit_control/models/credit_control_line.py
qgroulard: Not relevant anymore, this part of code has been removed for simplification.

account_credit_control/models/credit_control_policy.py
qgroulard: Not relevant anymore, this part of code has been removed for simplification.

account_credit_control/models/credit_control_line.py
qgroulard: Indeed, thank you.

account_credit_control/models/credit_control_line.py
qgroulard: Indeed, thank you.

account_credit_control/models/credit_control_policy.py
qgroulard: Field removed.

account_credit_control/models/credit_control_policy.py
qgroulard: Not relevant anymore, this part of code has been removed for simplification.

account_credit_control/models/credit_control_line.py
qgroulard: I added 'exclude_ids' arg to _get_related_lines().

account_credit_control/models/credit_control_line.py
qgroulard: Done.

account_credit_control/models/credit_control_line.py
qgroulard: Done.

account_credit_control/models/credit_control_line.py
qgroulard: I added a level criteria and refactored this part a bit.

account_credit_control/models/credit_control_line.py
qgroulard: :+1:

account_credit_control/models/credit_control_line.py
qgroulard: Done.

account_credit_control/models/credit_control_line.py
qgroulard: Done as well as _get_lower_related_lines() and _get_related_lines

PR #30, opened 2019-10-08
[12.0][ENH] Credit Control Report Odoo 12 style

Martronic-SA:
Go closer to the default Odoo 12 report style used in invoice document report.
The aim is to have a better harmonization between reports.
That PR also fixes the total table not going on right when printing.

### DIFF
--- a/account_credit_control/report/report_credit_control_summary.xml
+++ b/account_credit_control/report/report_credit_control_summary.xml
@@ -2,23 +2,22 @@
 <odoo>
     <template id="report_credit_control_summary_document">
         <t t-call="web.external_layout">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Screenshots of before/after?

Martronic-SA:
I'm having this before:

and after:

pedrobaeza:
I think the previous idea is to put that as a letter, so not sure about the change...

Martronic-SA:
I think an invoice is a letter too.
But that PR fixes the following problems in the existing report:
Customer name and date are sticked on the left
Total amounts are set in the report but they disappear when printing because of the use of wrong css class

pedrobaeza (approved):

qgroulard (approved):
LGTM (code review + functional test)

PR #33, opened 2019-10-14
[12.0] [IMP] Improve create_or_update_from_mv_lines performances

qgroulard:
The new index and the create and write multi significantly improve performances when it comes to generate hundreds of credit control lines.

### DIFF
--- a/account_credit_control/models/credit_control_line.py
+++ b/account_credit_control/models/credit_control_line.py
@@ -162,19 +162,21 @@ class CreditControlLine(models.Model):
     def _prepare_from_move_line(self, move_line, level, controlling_date,
                                 open_amount):
         """ Create credit control line """
-        data = {}
-        data['date'] = controlling_date
-        data['date_due'] = move_line.date_maturity
-        data['state'] = 'draft'
-        data['channel'] = level.channel
-        data['invoice_id'] = (move_line.invoice_id.id if
-                              move_line.invoice_id else False)
-        data['partner_id'] = move_line.partner_id.id
-        data['amount_due'] = (move_line.amount_currency or move_line.debit or
-                              move_line.credit)
-        data['balance_due'] = open_amount
-        data['policy_level_id'] = level.id
-        data['move_line_id'] = move_line.id
+        data = {
+            'date': controlling_date,
+            'date_due': move_line.date_maturity,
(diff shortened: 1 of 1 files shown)


account_credit_control/models/credit_control_line.py
```diff
@@ -232,10 +234,11 @@ def create_or_update_from_mv_lines(self, lines, level, controlling_date,
                 ('move_line_id', '=', move_line.id),
                 ('policy_level_id', '=', level.id),
                 ('state', '=', 'draft'),
-                ('id', '!=', line.id),
             ])
-            if previous_drafts:
-                previous_drafts.write({'state': 'ignored'})
+            lines_to_write = lines_to_write | previous_drafts
+
+        new_lines = self.create(lines_to_create)
```
pedrobaeza, line 242: This line in `create` also affects a lot the performance:

https://github.com/OCA/credit-control/blob/01697bf3f5520e5c499b8cd769202cc6248c2aae/account_credit_control/models/credit_control_line.py#L268

Maybe a related field?

account_credit_control/models/credit_control_line.py
```diff
@@ -232,10 +234,11 @@ def create_or_update_from_mv_lines(self, lines, level, controlling_date,
                 ('move_line_id', '=', move_line.id),
                 ('policy_level_id', '=', level.id),
                 ('state', '=', 'draft'),
-                ('id', '!=', line.id),
             ])
-            if previous_drafts:
-                previous_drafts.write({'state': 'ignored'})
+            lines_to_write = lines_to_write | previous_drafts
```
daramousk, line 240: or `lines_to_write |= previous_drafts` perhaps

daramousk (approved):
Did some profiling with `tools.profiler`, without this patch I got 62 ms and with it 59. Looks great to me.

account_credit_control/models/credit_control_line.py
qgroulard: Not quite sure it can be related, but what about https://github.com/OCA/credit-control/pull/33/commits/d9a95c5f64bed1f05d6ea1ece9bdf63fa52e2816 ?

pedrobaeza (approved):

PR #35, opened 2019-10-15
[12.0] [ADD] Always compute option on Policy Level

qgroulard:
Add an option on Policy Level to generate lines of this level even if lower level lines have not be generated or are still in draft. Thus, for instance, a credit control line of level 3 might be generated from a line of level 1, bypassing 2, because conditions of the 3rd level are met.

Use Case: Our customer considers that even if he forgets to process the credit control line and send the reminder to his customer, if his customer is very late higher level reminder should be send rather than lower one.


sbidoul:
In the end we don't use this. Since it is difficult to explain and understand, let's close it.

PR #37, opened 2019-10-21
[12.0] [IMP] Improve usability

qgroulard:
From: https://github.com/OCA/account-financial-tools/pull/819

- Add 'phone' channel
- Therefore rename 'Ready To Send' -> 'To Do', 'Sent' -> 'Done' and 'Date Sent' -> 'Date Reminded'
- track_visibility="on_change" on credit control line state
- Display level_lines max level when grouped (groupby in tree view) instead of sum
- Fallback to letter if channel is 'email' and partner has no email
- Fix error message when trying to create a run at a previous date than last run


account_credit_control/models/credit_control_run.py
```diff
@@ -100,17 +100,18 @@ def _check_run_date(self, controlling_date):
         )
         if runs:
             raise UserError(
-                _('A run has already been executed more '
-                  'recently than %s') % runs.date)
+                _('A run has already been executed more recently (%s)')
+                % (runs.date)
```
sbidoul, line 104:
```suggestion
                % (runs.date,)
```

The comma is important for a 1-uple, as `(x)` is the same as `x`.

account_credit_control/models/credit_control_run.py
```diff
@@ -100,17 +100,18 @@ def _check_run_date(self, controlling_date):
...
         lines = line_obj.search([('date', '>', controlling_date)],
                                 order='date DESC', limit=1)
         if lines:
             raise UserError(
-                _('A credit control line more '
-                  'recent than %s exists at %s') % (
-                    controlling_date, lines.date))
+                _('A credit control line more recent than %s exists at %s')
+                % (controlling_date, lines.date)
+            )
```
sbidoul, line 114: nit: it looks like nothing changes here, so the diff could be reduced.

sbidoul (approved):
A small remark, otherwise LGTM (code review).

account_credit_control/models/credit_control_run.py
sbidoul: But since `runs.date` is never a tuple nor a list, it does not matter.

sbidoul:
@gva-acsone, is this one good for you? Anyone else?

gva-acsone (approved):
Functional test LGTM

PR #40, opened 2019-10-30
[12.0] [IMP] move address on report_credit_control_summary

qgroulard:
Before/after:

image

image


pedrobaeza:
Isn't this the same as #30 ?

qgroulard:
@pedrobaeza  Yes indeed, let's close this one.

sbidoul:
@qgroulard can you then review #30?

PR #41, opened 2019-11-14
[WIP][FIX]account_financial_risk: risk_exception field explicitly setted to False

jesusVMayor:
In some cases, if the field is not setted to False explicitly, odoo raises a CacheMiss exception. I found the problem with the sale_order_revision module and some custom code who uses this field in the write function of sale.order, when a new revision is created, and the write try to take the field, it isn't in the cache.

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -210,16 +210,18 @@ class ResPartner(models.Model):
         risk_field_list = self._risk_field_list()
         for partner in self.filtered('customer'):
             amount = 0.0
+            risk_exception = False
             for risk_field in risk_field_list:
                 field_value = getattr(partner, risk_field[0], 0.0)
                 max_value = getattr(partner, risk_field[1], 0.0)
                 if max_value and field_value > max_value:
-                    partner.risk_exception = True
+                    risk_exception = True
                 if getattr(partner, risk_field[2], False):
                     amount += field_value
             partner.risk_total = amount
             if partner.credit_limit and amount > partner.credit_limit:
-                partner.risk_exception = True
+                risk_exception = True
+            partner.risk_exception = risk_exception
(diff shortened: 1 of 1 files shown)


account_financial_risk/models/res_partner.py
```diff
@@ -215,11 +215,15 @@ def _compute_risk_exception(self):
                 max_value = getattr(partner, risk_field[1], 0.0)
                 if max_value and field_value > max_value:
```
pedrobaeza, line 217: Please do it more Pythonic :wink: :

```suggestion
                partner.risk_exception = max_value and field_value > max_value
```

account_financial_risk/models/res_partner.py
```diff
@@ -215,11 +215,15 @@ def _compute_risk_exception(self):
...
                     partner.risk_exception = True
+                else:
+                    partner.risk_exception = False
                 if getattr(partner, risk_field[2], False):
                     amount += field_value
             partner.risk_total = amount
             if partner.credit_limit and amount > partner.credit_limit:
```
pedrobaeza, line 222: Please do it more Pythonic :wink: :

```suggestion
                partner.risk_exception = partner.credit_limit and amount > partner.credit_limit
```

pedrobaeza (changes requested):

pedrobaeza:
But you have to remove the rest of the lines... and please squash commits into one.

jesusVMayor:
yes... i need to take a coffee before making a PR :), this change also breaks the compute.

pedrobaeza:
Yes, seeing the logic, you only need to put `partner.risk_exception = False` at the beginning of the loop and nothing more.

pedrobaeza (approved):

carlosdauden (approved):

PR #43, opened 2019-11-18
[12.0] [REF] Refactor generate_credit_lines for extensibility

qgroulard:


account_credit_control/models/credit_control_policy.py
```diff
@@ -201,6 +201,35 @@ def check_policy_against_account(self, account):
             )
         return True
 
+    @api.multi
+    def generate_credit_lines(self, run):
```
sbidoul: I would pass the controlling date, so the policy object does not need to know what a run is.

sbidoul: We'll see in another PR how to optimize/handle the link between credit control lines and runs.

sbidoul (approved):
Except the run argument, this simple refactoring is good.

it opens the possibility to customize the line generation algorithm per policy.

account_credit_control/models/credit_control_policy.py
sbidoul: I think passing the controlling data and a dictionary of default values for new lines should be a good approach. So you can pass `{'run_id': ...}`.

sbidoul (approved):

account_credit_control/models/credit_control_policy.py
```diff
@@ -201,6 +201,33 @@ def check_policy_against_account(self, account):
             )
         return True
 
+    @api.multi
+    def generate_credit_lines(self, controlling_date):
```
sbidoul:
```suggestion
    def _generate_credit_lines(self, controlling_date):
```

account_credit_control/models/credit_control_line.py
```diff
@@ -233,7 +248,12 @@ def create_or_update_from_mv_lines(self, lines, level, controlling_date,
             if check_tolerance and open_amount < cur_tolerance:
                 continue
             vals = self._prepare_from_move_line(
-                move_line, level, controlling_date, open_amount)
+                move_line,
+                level,
+                controlling_date,
+                open_amount,
+                default_lines_vals,
```
sbidoul:
```suggestion
                default_lines_vals or {},
```

account_credit_control/wizard/credit_control_policy_changer.py
```diff
@@ -123,6 +123,7 @@ def set_new_policy(self):
             self.move_line_ids,
             self.new_policy_level_id,
             controlling_date,
+            {},
```
sbidoul:
```suggestion
```

account_credit_control/models/credit_control_line.py
```diff
@@ -167,28 +167,41 @@ def _compute_partner_user_id(self):
...
     @api.model
-    def create_or_update_from_mv_lines(self, lines, level, controlling_date,
-                                       check_tolerance=True):
+    def create_or_update_from_mv_lines(
+        self,
+        lines,
+        level,
+        controlling_date,
+        default_lines_vals,
+        check_tolerance=True,
```
sbidoul, line 210: add `default_lines_vals=None` as last argument for backward compatibility

sbidoul (changes requested):
LGTM, but a small change for backward compatibility.

account_credit_control/models/credit_control_line.py
sbidoul: @qgroulard the default_lines_vals must be added *after* check_tolerance, in case someone was calling the method without keyword arguments.

sbidoul (approved):

sbidoul:
This should keep the code fully compatible, so 
/ocabot merge patch

PR #46, opened 2020-01-27
[12.0][FIX] account_financial_risk: consider sign in financial risk draft invoices

dsidorenkovas-rgbconsulting:
**Description of the issue this PR addresses**
Refund invoices in draft state are not considered as negative amounts in the customer's financial risk fields.

**Steps to reproduce:**
- Create a  new customer.
- Create a refund invoice (credit note) for the customer.
- Go to the customer's financial risk tab.

**Current behavior before PR:**
- The amount of draft invoices is positive.

**Desired behavior after PR is merged:**
- The amount of draft invoices is negative.

### DIFF
--- a/account_financial_risk/README.rst
+++ b/account_financial_risk/README.rst
@@ -87,6 +87,10 @@ Contributors
 
 * Agathe Mollé <agathe.molle@savoirfairelinux.com>
 
+* `RGB Consulting <https://www.rgbconsulting.com>`_:
+
+  * Domantas Sidorenkovas
+
 Maintainers
 ~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


gdgellatly (commented):
I'm not sure simple bugfixes are contributors now. but the rest looks good. I let PSC decide. My only question is does this upgrade require triggering a recompute.

dsidorenkovas-rgbconsulting:
> I'm not sure simple bugfixes are contributors now. but the rest looks good. I let PSC decide. My only question is does this upgrade require triggering a recompute.

There is no problem with removing the contributor if needed, I just added it for convention.

Regarding the computed field, yes, the value needs to be recalculated, it can be done simply from shell, but if you consider it necessary I can prepare a version upgrade script for it.

carlosdauden:
It's improved in https://github.com/OCA/credit-control/pull/65

PR #55, opened 2020-02-26
[FIX] sale_financial_risk: compute sudo

chienandalu:
STR: When the user has not acl to sale.order.line and tries to create a partner, a permissions error will raise. 

We prevent it sudoing the compute.

cc @Tecnativa TT20992


chienandalu:
ping @carlosdauden @pedrobaeza

pedrobaeza (approved):

carlosdauden (approved):

pedrobaeza:
/ocabot merge patch

Is this happening as well on v12 (and v13 PR)?

PR #52, opened 2020-02-08
[MIG][13.0] account_financial_risk: Migration to v13.0

sergio-teruel:
cc @Tecnativa TT21602
@carlosdauden @rafaelbn review please

PR #56, opened 2020-02-27
[IMP] account_financial_risk: Avoid compute all partners saving config settings

carlosdauden:
This module adds a field in config settings, this field is allways saved when we click save button even we are in other section and we don't change any option.

With this PR I try avoid this trigger if don't change the value:
https://github.com/OCA/credit-control/blob/bf46dd5d299106973381858c731ab9a3d255d301/account_financial_risk/models/res_partner.py#L166

@Tecnativa TT22541

### DIFF
--- a/account_financial_risk/models/res_company.py
+++ b/account_financial_risk/models/res_company.py
@@ -13,3 +13,12 @@ class ResCompany(models.Model):
              "The change of this field recompute all partners risk,"
              "be patient.",
     )
+
+    # TODO: Remove when odoo compare values before to write
+    def write(self, vals):
+        new_margin = vals.get('invoice_unpaid_margin')
+        if (new_margin is not None and
+                all(new_margin == x.invoice_unpaid_margin for x in self)):
+            vals = vals.copy()
+            vals.pop('invoice_unpaid_margin')
+        return super().write(vals)


account_financial_risk/models/res_company.py
```diff
@@ -13,3 +13,11 @@ class ResCompany(models.Model):
              "The change of this field recompute all partners risk,"
              "be patient.",
     )
+
+    # Remove when odoo compare values before to write
+    def write(self, vals):
+        new_margin = vals.get('invoice_unpaid_margin')
+        if (new_margin is not None and
+                new_margin == self[:1].invoice_unpaid_margin):
```
pedrobaeza: You should check for each of the written companies. `self` is usually one, but no one knows...:
```suggestion
                all([new_margin == x.invoice_unpaid_margin for x in self]):
```

carlosdauden: Done

pedrobaeza (approved):

account_financial_risk/models/res_company.py
```diff
@@ -13,3 +13,11 @@ class ResCompany(models.Model):
              "The change of this field recompute all partners risk,"
              "be patient.",
     )
+
+    # Remove when odoo compare values before to write
```
yajo:
```suggestion
    # TODO Remove when odoo compare values before to write
```

account_financial_risk/models/res_company.py
```diff
@@ -13,3 +13,11 @@ class ResCompany(models.Model):
...
+    def write(self, vals):
+        new_margin = vals.get('invoice_unpaid_margin')
+        if (new_margin is not None and
+                all([new_margin == x.invoice_unpaid_margin for x in self])):
```
yajo:
```suggestion
                all(new_margin == x.invoice_unpaid_margin for x in self)):
```

rafaelbn (approved):

account_financial_risk/models/res_company.py
```diff
@@ -13,3 +13,11 @@ class ResCompany(models.Model):
...
+            vals.pop('invoice_unpaid_margin')
```
yajo, line 23: It's usually not a good practice to alter an incoming parameter:

```suggestion
            vals = dict(vals)
            vals.pop('invoice_unpaid_margin')
```

yajo (changes requested):
Minor details.

carlosdauden:
@Yajo changes done.

yajo (approved):

pedrobaeza:
Is this needed for v12 and v13?

carlosdauden:
In v12 is needed ¿can you do cherry-pick?

pedrobaeza:
Cherry-picked for v12 at 875093ce6f15721c047cb3b1f8575aebdb9ebe19

What about v13?

carlosdauden:
v13 have better performance, but the compute is triggered too.
Can you do cherry-pick or black is a problem?

carlosdauden:
Included in PR:
https://github.com/OCA/credit-control/pull/52/files#diff-24929d0eb5a16ed964b08808f0614745R17-R24

PR #60, opened 2020-03-23
[12.0][FIX] sale_financial_risk: Duplicate amount on risk_sale_order and risk_invoice_draft

carlosdauden:
### Before PR:

Confirmed SO/001 with total amount 100€:
risk_sale_order = 100€
risk_invoice_draft = 0
risk_invoice_open = 0

After create a draft invoice from SO/001:
**risk_sale_order = 100€
risk_invoice_draft = 100€**
risk_invoice_open = 0

### After this PR:

Confirmed SO/001 with total amount 100€:
risk_sale_order = 100€
risk_invoice_draft = 0
risk_invoice_open = 0

After create a draft invoice from SO/001:
**risk_sale_order = 0€
risk_invoice_draft = 100€**
risk_invoice_open = 0

@Tecnativa

### DIFF
--- a/sale_financial_risk/models/sale.py
+++ b/sale_financial_risk/models/sale.py
@@ -68,7 +68,7 @@ class SaleOrderLine(models.Model):
         """
         for line in self.filtered(lambda l: l.state == 'sale'):
             invoice_lines = line.invoice_lines.filtered(
-                lambda l: l.invoice_id.state in {'open', 'in_payment', 'paid'})
+                lambda l: l.invoice_id.state != 'cancel')
             amount_invoiced = 0.0
             for inv_line in invoice_lines:
                 inv_date = (inv_line.invoice_id.date_invoice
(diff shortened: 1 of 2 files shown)


sale_financial_risk/models/sale.py
```diff
@@ -68,7 +68,7 @@ def _compute_amount_to_invoice(self):
         """
         for line in self.filtered(lambda l: l.state == 'sale'):
             invoice_lines = line.invoice_lines.filtered(
-                lambda l: l.invoice_id.state in {'open', 'in_payment', 'paid'})
+                lambda l: l.invoice_id.state != 'cancel')
```
pedrobaeza, line 71: Are you including `draft` invoices then?

carlosdauden: Yes. Updated PR description.

pedrobaeza:
Thanks for the extra info. I understand the logic, but now I see we have the risk of some amounts not being taking into account. What if you don't have `draft invoices` included in the risk, should this amount be substracted from sales order one? And talking about that, what are the defaults we are providing for these checks? All unmarked?

manuelcalerosolis (approved):
Functionality tested ok!

gdgellatly (approved):
Boy I got confused with this. Finally got it. Static Review by long time user

rafaelbn (approved):
Tested :+1:

pedrobaeza:
Can anybody answer https://github.com/OCA/credit-control/pull/60#issuecomment-603110318 ?

chienandalu (approved):

chienandalu:
I hadn't read @pedrobaeza https://github.com/OCA/credit-control/pull/60#issuecomment-603110318 It's a good question

PR #62, opened 2020-04-23
[IMP] sale_financial_risk: Bypass risk when payment transaction is authorized

carlosdauden:
@Tecnativa TT23520

### DIFF
--- a/sale_financial_risk/models/payment.py
+++ b/sale_financial_risk/models/payment.py
@@ -1,0 +1,20 @@
+# Copyright 2020 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class PaymentTransaction(models.Model):
+    _inherit = 'payment.transaction'
+
+    def _set_transaction_authorized(self):
+        """Bypass risk for sale confirmation triggered by this method"""
+        return super(PaymentTransaction, self.with_context(
+            bypass_risk=True))._set_transaction_authorized()
+
+    def _reconcile_after_transaction_done(self):
+        """Bypass risk for sale confirmation and invoice creation triggered
+        by this method
+        """
+        return super(PaymentTransaction, self.with_context(
+            bypass_risk=True))._reconcile_after_transaction_done()
--- a/sale_financial_risk/models/__init__.py
+++ b/sale_financial_risk/models/__init__.py
@@ -1,2 +1,3 @@
-from . import sale
+from . import payment
 from . import res_partner
+from . import sale


pedrobaeza (commented):
I miss the proper docstring on each method with the why you are passing that context. I wonder indeed why you need to overwrite 2 methods.

carlosdauden:
We can overwrite those methods or this methods:
https://github.com/odoo/odoo/blob/a8689769afa22fb633f2e911f1a4dea6caa193d3/addons/sale/models/payment.py#L99
https://github.com/odoo/odoo/blob/a8689769afa22fb633f2e911f1a4dea6caa193d3/addons/sale/models/payment.py#L109

Or we can overwrite this methods:
https://github.com/odoo/odoo/blob/a8689769afa22fb633f2e911f1a4dea6caa193d3/addons/sale/models/payment.py#L82
https://github.com/odoo/odoo/blob/a8689769afa22fb633f2e911f1a4dea6caa193d3/addons/sale/models/payment.py#L138

As it is an innocuous context, I have decided to overwrite the most generic methods

pedrobaeza:
But why 2 and not 1?

carlosdauden:
Have you opened the links that I have put?

pedrobaeza:
Yes, and I need a plain explanation, and it's what I'm requesting for docstrings. Having obscure overwrites or temporary links lead to not being sure when migrating this what to do.

carlosdauden:
The confirmation of orders can come in two ways.
Invoice confirmation comes from one of those two ways.

bypass_risk in authorized transaction is pretty self explanatory, but you can make a suggestion for those docstrings and I will add them.

rafaelbn (approved):
Thanks :+1:

sale_financial_risk/models/payment.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class PaymentTransaction(models.Model):
+    _inherit = 'payment.transaction'
+
+    def _set_transaction_authorized(self):
```
chienandalu, line 10:
```suggestion
    def _set_transaction_authorized(self):
        """Bypass risk for sale confirmation triggered by this method"""
```

sale_financial_risk/models/payment.py
```diff
@@ -0,0 +1,16 @@
...
+        return super(PaymentTransaction, self.with_context(
+            bypass_risk=True))._set_transaction_authorized()
+
+    def _reconcile_after_transaction_done(self):
```
chienandalu, line 15:
```suggestion
    def _reconcile_after_transaction_done(self):
        """Bypass risk for sale confirmation and invoice creation triggered by this method"""
```

chienandalu (approved):

carlosdauden:
Changes done

PR #63, opened 2020-04-27
[11.0][FIX] *_risk: adapt to multicurrency

MiquelRForgeFlow:
We can't sum and compare amounts of different currencies.

### DIFF
--- a/sale_financial_risk/models/sale.py
+++ b/sale_financial_risk/models/sale.py
@@ -40,10 +40,17 @@ class SaleOrder(models.Model):
 class SaleOrderLine(models.Model):
     _inherit = 'sale.order.line'
 
+    company_currency_id = fields.Many2one(
+        comodel_name='res.currency',
+        related='company_id.currency_id',
+        string="Company Currency",
+        readonly=True,
+    )
     risk_amount = fields.Monetary(
         string="Risk amount",
         compute='_compute_risk_amount',
         compute_sudo=True,
+        currency_field="company_currency_id",
         store=True
     )
     # TODO: Analyze performance vs order_id.partner_id.commercial_partner_id


MiquelRForgeFlow:
@carlosdauden @tecnativa @JordiBForgeFlow

carlosdauden:
Can you check if this solves your problem?:
https://github.com/OCA/credit-control/pull/66

MiquelRForgeFlow:
> Can you check if this solves your problem?

Checked, and no, it doesn't solve the problem. Basically, in multicurrency, you may have move lines with amount_residual, and others with amount_residual_currency, depending if they are done with the same currency or not. I think maybe it would be better if you merge the other PR and then I rebase this one.

carlosdauden:
But I think that risk should always be controlled in the currency of the company, because otherwise it would be a mess for the accountant.
Imagine having to establish the risk for each client in a different currency and the amounts appear in that currency ...
amount_residual field is allways computed in the currency of the company.

JordiBForgeFlow:
Why a mess? We have this use case. If the customers pays you always in USD you may want to control the customer credit risk in that currency.  Is something that you want to control at customer level.

carlosdauden:
Of course, because it is only a currency and also USD ...
But imagine the following and that the policy of the company is to give by default a credit of 1000€

EUR / USD 1,1306
EUR / GBP 0.8973
EUR / JPY 120.8100
EUR / BRL 5.9765
EUR / MXN 30.7325

Do you think it is easy for the risk manager to manage the information in the different currencies of the customers?

And I've only put 5 common currencies...

carlosdauden:
One possibility could be to establish a configuration parameter to decide one behavior or another in each company...

MiquelRForgeFlow:
> One possibility could be to establish a configuration parameter to decide one behavior or another in each company...

Good, then I will close this PR, and then if you are ok, I will try this approach you propose in v13 PR.

MiquelRForgeFlow:
@carlosdauden sorry, the rebase post-improvement was done wrong. Can you check again? I think it's clear now what I am trying to fix.

account_financial_risk/tests/test_account_financial_risk.py
```diff
@@ -13,6 +13,8 @@ def setUpClass(cls):
         type_revenue = cls.env.ref('account.data_account_type_revenue')
         type_receivable = cls.env.ref('account.data_account_type_receivable')
         tax_group_taxes = cls.env.ref('account.tax_group_taxes')
+        main_company = cls.env.ref('base.main_company')
+        main_company.currency_id = cls.env.ref('base.USD')
```
pedrobaeza: This triggers a constraint, so it can't be done this way

MiquelRForgeFlow: Maybe done with a query?

pedrobaeza: First the convenience of this patch should be determined before "patching" other things

JordiBForgeFlow:
@carlosdauden @pedrobaeza can we merge this?

pedrobaeza:
For me it's not OK. You don't have any currency linked to a partner, so you can't set that limit on any other currency that is not the company's currency. If you want this, please extract it to other module, and add a currency at partner level for controlling the risk in that currency. For me it's also a mess to mix all of this, but anyway, it's totally out of the scope of the main module.

:-1: for merging this.

MiquelRForgeFlow:
@pedrobaeza that was before my last force push. I changed that. I maintain now the company currency. In this PR I am just fixing the cases where the account move lines are in a different currency, taken that into account, and add them in a proper way just to avoid mixing apples and oranges.

pedrobaeza:
But `amount_residual` isn't always in company's currency?

MiquelRForgeFlow:
Yes, but then some times, when working in different currency, the `amount_residual_currency` is filled instead and the `amount_residual` is empty.

pedrobaeza:
Which are that "some times"? Steps to reproduce that problem with core addons? If not, you need to fix the modules that create that account moves in that way, or if it's something in core, fix it there (although I know that this case is harder, and we can admit the patch if there's a situation with core that provokes this problem).

MiquelRForgeFlow:
Hmmm... I have not found, maybe was a bug that got fixed.

MiquelRForgeFlow:
Well, I changed the commit now. Just a minor thing left.

pedrobaeza (approved):

carlosdauden (approved):
We should make this change in v12 and v13

PR #64, opened 2020-05-04
[IMP] Report custom text fields in html

qgroulard:
The aim of these two fields is to be injected in reports, thus it seems more convenient to me to have them in HTML.

### DIFF
--- a/account_credit_control/__manifest__.py
+++ b/account_credit_control/__manifest__.py
@@ -3,7 +3,7 @@
 # Copyright 2018 Access Bookings Ltd (https://accessbookings.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {'name': 'Account Credit Control',
- 'version': '12.0.1.6.0',
+ 'version': '12.0.1.7.0',
  'author': "Camptocamp,"
            "Odoo Community Association (OCA),"
            "Okia,"
--- a/account_credit_control/migrations/12.0.1.7.0/post-migration.py
+++ b/account_credit_control/migrations/12.0.1.7.0/post-migration.py
@@ -1,0 +1,30 @@
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """
+    Convert custom_text fields to Html on policy_level
+    """
+    cr = env.cr
+    openupgrade.copy_columns(
+        cr,
+        {
+            "credit_control_policy_level": [
+                ("custom_text", None, None),
+                ("custom_text_after_details", None, None),
+            ]
+        },
+    )
+    openupgrade.convert_field_to_html(
+        cr, "credit_control_policy_level", "custom_text", "custom_text"
+    )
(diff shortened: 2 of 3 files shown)


pedrobaeza:
You need to add a migration script to convert the column to HTML then. There's a method in openupgradelib for that:

https://github.com/OCA/openupgradelib/blob/6c18ce0c160b9cf47c4fa898d9b6c4265d98c216/openupgradelib/openupgrade.py#L1912

qgroulard:
> You need to add a migration script to convert the column to HTML then. There's a method in openupgradelib for that:
> 
> https://github.com/OCA/openupgradelib/blob/6c18ce0c160b9cf47c4fa898d9b6c4265d98c216/openupgradelib/openupgrade.py#L1912

Column type of Html fields is `text`.
I believe Html and Text fields are stored the same way in DB, thus no migration is needed. Am I wrong ?

pedrobaeza:
Not really... `field.Text` is only allowing multi-line, but they are stored as Python texts, so you need to convert `\n` to `<br>` through that method. If not, you will find that the lines are put all together in the same paragraph.

account_credit_control/migrations/12.0.1.6.1/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """
+    Convert custom_text fields to Html on policy_level
+    """
+    _logger.info(">> Post-Migration 12.0.1.6.1")
```
pedrobaeza: No need of this. It's automatically logged by openupgradelib.

account_credit_control/migrations/12.0.1.6.1/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    cr = env.cr
+    openupgrade.convert_field_to_html(
+        cr, "credit_control_policy_level", "custom_text", "custom_text"
```
pedrobaeza: Please do a copy of the column first.

pedrobaeza (changes requested):
Please update minor version instead of patch for being semantically correct, and squash commits.

account_credit_control/migrations/12.0.1.6.1/post-migration.py
qgroulard: What would be the purpose of this copied column ?
Is it just to keep a text version of the data in DB in case we want to convert it back ?

pedrobaeza: Yes, it's a good migration practice to keep previous data unaltered (we do it constantly on OpenUpgrade), in case you need to compare, revert back, re-apply...

I forgot to mention that you have other openupgradelib method that eases this and you can use `None, None` for the rest of the arguments, sorry:

https://github.com/OCA/openupgradelib/blob/6c18ce0c160b9cf47c4fa898d9b6c4265d98c216/openupgradelib/openupgrade.py#L348

qgroulard: Ok I understand, thank you.

pedrobaeza (approved):

sbejaoui (approved):

PR #65, opened 2020-05-28
[12.0][IMP] account_financial_risk sale_financial_risk: Performance. Not store fields. Links to values records. Multi-company. Multi-currency.

carlosdauden:
***_financial_risk:** (account_financial_risk, account_payment_return_financial_risk sale_financial_risk)
 - Optimize code and change stored computed fields to not store to improve multi-company support
 - Improve multi-currency support
 - Convert risk amount fields to clickable link that shows traceability of amount origin
 - New pivot views to risk amount traceability
 - Simplify class style applied on risk fields
 - Migration script to remove old stored computed fields
 - Improve tests to cover new functionallity
 - Update translation files

**account_financial_risk:**
 - Don't block refund invoice validation when partner has risk exception
 - Allow search partners by risk exception field
 - Remove obsolete cron

**sale_financial_risk**
 - Create related store commercial_partner_id field in sale order line to simplify computation
 - Rename amt_to_invoice field to risk_amount in sale order line
 - Hook and migration scripts to reduce new fields computing time

@Tecnativa TT23765

financial_risk_click

PR #67, opened 2020-06-08
[13.0][MIG] account_financial_risk

MiquelRForgeFlow:
Standard migration.


pedrobaeza:
This is not including #65

pedrobaeza:
Sorry, I have seen it now

carlosdauden:
Migrating this module in https://github.com/OCA/credit-control/pull/52

pedrobaeza:
Maybe it's better to take this one that includes all your latest changes. That one is even red.

carlosdauden:
:+1:

account_financial_risk/models/res_partner.py
```diff
@@ -123,7 +121,7 @@ def _compute_risk_allow_edit(self):
 
     @api.model
     def _get_risk_company_domain(self):
-        return [("company_id", "=", self.env.user.company_id.id)]
+        return [("company_id", "=", self.env.company.id)]
```
carlosdauden:
```suggestion
        return [("company_id", "in", self.env.companies.ids)]
```

account_financial_risk/models/res_partner.py
```diff
@@ -152,16 +150,16 @@ def _get_field_risk_model_domain(self, field_name):
...
         "invoice_ids",
         "invoice_ids.state",
-        "invoice_ids.amount_total_company_signed",
+        "invoice_ids.amount_total_signed",
         "child_ids.invoice_ids",
         "child_ids.invoice_ids.state",
-        "child_ids.invoice_ids.amount_total_company_signed",
+        "child_ids.invoice_ids.amount_total_signed",
     )
     def _compute_risk_invoice(self):
```
carlosdauden: You can remove this method adding "draft" key in _risk_account_groups()

carlosdauden (changes requested):
Minor suggestions

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: but the read_group for draft runs in account.move, not in account.move.line 🤔

carlosdauden: My idea was to filter with:
`('exclude_from_invoice_tab', '=', False),`
`('parent_state', '=', 'draft'),`

You can have a look?
If you prefer we can merge this and I'll do the corresponding PR.

MiquelRForgeFlow: any condition in the type of the move? or all types are valid?

carlosdauden: Of course, the same condition that is now:
` ("move_id.type", "in", ["out_invoice", "out_refund"])`

https://github.com/OCA/credit-control/pull/67/files?file-filters%5B%5D=.py&file-filters%5B%5D=.xml#diff-16fd100e0e996b54789c187c889dae83R130

MiquelRForgeFlow: could you check last commit?

MiquelRForgeFlow: As far I can see, doing the draft with the move lines instead of the account move means that the tax won't be included in the draft amount.

carlosdauden: Sorry, instead of 
`('exclude_from_invoice_tab', '=', False)`

Use:
`('account_internal_type', '=', 'receivable')`

account_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,379 @@
...
+                    ("move_id.type", "in", ["out_invoice", "out_refund"]),
+                    ("account_internal_type", "=", "receivable"),
+                    ("parent_state", "in", ["draft", "proforma", "proforma2"]),
+                ],
+                "fields": [
+                    "partner_id",
+                    "account_id",
+                    "debit",
+                    "credit",
+                    "tax_base_amount",
```
carlosdauden:
```suggestion
                    "amount_residual",
```

account_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,379 @@
...
+                ],
+                "fields": ["partner_id", "account_id", "amount_residual"],
+                "group_by": ["partner_id", "account_id"],
+            },
+        }
+
+    @api.depends(
+        "move_line_ids.debit",
+        "move_line_ids.credit",
+        "move_line_ids.tax_base_amount",
```
carlosdauden:
```suggestion

```

account_financial_risk/models/res_company.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # TODO: Remove when odoo compare values before to write
+    def write(self, vals):
+        new_margin = vals.get("invoice_unpaid_margin")
+        if new_margin is not None and all(
+            new_margin == x.invoice_unpaid_margin for x in self
+        ):
+            vals = vals.copy()
+            vals.pop("invoice_unpaid_margin")
+        return super().write(vals)
```
carlosdauden: Maybe it could be removed because now the change should not cause heavy recalculations.
```suggestion
```

carlosdauden (approved):

account_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,368 @@
...
+            else:
+                vals["risk_account_amount_unpaid"] += reg["amount_residual"]
+        return vals
+
+    @api.depends(lambda x: x._get_depends_compute_risk_exception())
+    def _compute_risk_exception(self):
+        risk_field_list = self._risk_field_list()
+        for partner in self:
+            if not partner.customer_rank > 0:
+                partner.risk_exception = False
```
carlosdauden:
```suggestion
                partner.risk_total = 0.0
                partner.risk_exception = False
```

carlosdauden (changes requested):
¿Can you add the last suggestion?
:wink:

account_financial_risk/views/res_partner_view.xml
```diff
@@ -0,0 +1,149 @@
...
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="priority" eval="20" />
+        <field name="groups_id" eval="[(4, ref('account.group_account_manager'))]" />
+        <field name="arch" type="xml">
+            <page name="sales_purchases" position="after">
+                <page
+                    name="financial_risk"
+                    string="Financial Risk"
+                    attrs="{'invisible': ['|', ('customer_rank', '>', 0), '&amp;', ('is_company','=',False), ('parent_id','!=',False)]}"
```
carlosdauden:
```suggestion
                    attrs="{'invisible': ['|', ('customer_rank', '=', 0), '&amp;', ('is_company','=',False), ('parent_id','!=',False)]}"
```

carlosdauden (changes requested):

carlosdauden (approved):

account_financial_risk/models/res_company.py
```diff
@@ -19,13 +19,3 @@ class ResCompany(models.Model):
         "Useful when the flow comes from sales orders and the over-risk "
         "has already been allowed when confirming these.",
     )
-
-    # TODO: Remove when odoo compare values before to write
-    def write(self, vals):
```
pedrobaeza: I would say this is still happening in v13, isn't it, @carlosdauden ?

account_financial_risk/models/res_partner.py
```diff
@@ -219,14 +202,15 @@ def _risk_account_groups(self):
...
+                "risk_invoice_draft": 0.0,
                 "risk_invoice_open": 0.0,
                 "risk_invoice_unpaid": 0.0,
                 "risk_account_amount": 0.0,
                 "risk_account_amount_unpaid": 0.0,
             }
         )
         AccountMoveLine = self.env["account.move.line"].sudo()
-        customers = self.filtered(lambda x: x.customer and not x.parent_id)
+        customers = self.filtered(lambda x: x.customer_rank > 0 and not x.parent_id)
```
pedrobaeza: I wouldn't restrict to those with that rank, as you can have invoices and not being classified as customer.

account_financial_risk/models/res_partner.py
```diff
@@ -270,7 +258,8 @@ def _prepare_risk_account_vals(self, groups):
     def _compute_risk_exception(self):
         risk_field_list = self._risk_field_list()
         for partner in self:
-            if not partner.customer:
+            if not partner.customer_rank > 0:
```
pedrobaeza: The same about classification

account_financial_risk/models/res_partner.py
```diff
@@ -282,16 +271,16 @@ def _compute_risk_exception(self):
...
                 risk_exception = True
+            partner.risk_total = amount
             partner.risk_exception = risk_exception
 
     @api.model
     def _search_risk_exception(self, operator, value):
         commercial_partners = self.search(
             [
-                ("customer", "=", True),
+                ("customer_rank", ">", 0),
```
pedrobaeza, line 279: The same

pedrobaeza (changes requested):

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: Well, by default a partner has customer_rank = 0, right. But just when you have an invoice with this partner, then automatically customer_rank is updated and treat this partner as a customer.

pedrobaeza: And what about sales order risk? You should remove this part of the domain and compute it for all partners no matter their rank.

carlosdauden: When you create a partner from customer the rank is 1:
https://github.com/odoo/odoo/blob/0274d61a11c6f5220d6af37a38adafb3e5276224/addons/account/models/partner.py#L500-L506
https://github.com/odoo/odoo/blob/80dd40dd86484b3ab8c47b3baa315843f0027871/addons/account/views/partner_view.xml#L233

I think that It is the least bad solution, Any suggestion?

account_financial_risk/models/res_company.py
carlosdauden: Now there are'nt stored fields, so there should be no heavy re-computations

carlosdauden: After refactorization

carlosdauden: After refactorization. I'll remove v12 too.

account_financial_risk/models/res_partner.py
pedrobaeza: But what if you have created from other part (Contacts for example). For me, it should be:
```suggestion
        customers = self.filtered(lambda x: not x.parent_id)
```

carlosdauden: And what about suppliers?
Not to mention "weird" things like a partner with parent marked as a company...

pedrobaeza: What is the problem with "suppliers"? We are computing the risks. For partners without sales activity, the risk will be 0.

carlosdauden: That field is very easy to fix for rare cases.
The rest of workarrounds seem to me a worse solution.

pedrobaeza: Why? This is not going to be mass-computed for all partners AFAIK. Only entering on the partner form or checking risk for an operation, it will be computed. Is there's something I'm missing?

carlosdauden: _search_risk_exception
https://github.com/OCA/credit-control/pull/67#discussion_r440759340

pedrobaeza: OK, for that one can be considered, but in this case is not needed

carlosdauden: What about partner with parent marked as a company?

pedrobaeza: I don't get you, but is this related to `customer_rank`? If not, please keep discussions separated and let's handle this and then you open the other topic.

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: no, here is not the same because search

account_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,365 @@
...
+            {
+                "risk_invoice_draft": 0.0,
+                "risk_invoice_open": 0.0,
+                "risk_invoice_unpaid": 0.0,
+                "risk_account_amount": 0.0,
+                "risk_account_amount_unpaid": 0.0,
+            }
+        )
+        AccountMoveLine = self.env["account.move.line"].sudo()
+        customers = self.filtered(lambda x: not x.parent_id)
```
carlosdauden:
```suggestion
        customers = self.filtered(lambda p: p == p.commercial_partner_id)
```

carlosdauden (changes requested):

carlosdauden (approved):

pedrobaeza (approved):

PR #70, opened 2020-06-16
[11.0][IMP] Add selection: choose which currency should be used in risk

MiquelRForgeFlow:
Well, in https://github.com/OCA/credit-control/pull/63#issuecomment-642872888, @carlosdauden said that if we add a configuration parameter, then everybody can be happy in the way to show the currencies. Thus I added a `selection`, which let's choose between Company Currency, Receivable Currency, Pricelist Currency, and Manual Credit Currency. Note: Company Currency is the default. All risk values are in the risk currency, the one chosen.

Please, show your opinion.

### DIFF
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -1,11 +1,36 @@
 # Copyright 2016-2018 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import _, api, models
+from odoo import _, api, fields, models
 
 
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
+
+    company_currency_id = fields.Many2one(store=True)
+    risk_currency_id = fields.Many2one(
+        related="partner_id.risk_currency_id")
+    risk_amount_total_currency = fields.Monetary(
+        string="Risk Amount Total",
+        currency_field="risk_currency_id",
+        compute="_compute_risk_amount_total_currency")
+
+    @api.multi
+    @api.depends('invoice_line_ids.price_subtotal', 'tax_line_ids.amount',
+                 'tax_line_ids.amount_rounding', 'currency_id', 'company_id',
+                 'date_invoice', 'type',
+                 'partner_id.credit_currency',
+                 'partner_id.manual_credit_currency_id',
+                 'partner_id.property_account_receivable_id.currency_id',
+                 'partner_id.country_id',
+                 'company_id.currency_id')
+    def _compute_risk_amount_total_currency(self):
+        for invoice in self:
+            invoice.risk_amount_total_currency = \
+                invoice.company_currency_id.with_context(
+                    date=invoice.date_invoice or fields.Date.context_today(
+                        self)).compute(invoice.amount_total_company_signed,
+                                       invoice.risk_currency_id, round=False)
 
     def risk_exception_msg(self):
         self.ensure_one()
@@ -14,14 +39,14 @@ class AccountInvoice(models.Model):
         if partner.risk_exception:
             exception_msg = _("Financial risk exceeded.\n")
         elif partner.risk_invoice_open_limit and (
-                (partner.risk_invoice_open + self.amount_total_company_signed)
+                (partner.risk_invoice_open + self.risk_amount_total_currency)
                 > partner.risk_invoice_open_limit):
             exception_msg = _(
                 "This invoice exceeds the open invoices risk.\n")
         # If risk_invoice_draft_include this invoice included in risk_total
         elif not partner.risk_invoice_draft_include and (
                 partner.risk_invoice_open_include and
-                (partner.risk_total + self.amount_total_company_signed) >
+                (partner.risk_total + self.risk_amount_total_currency) >
                 partner.credit_limit):
             exception_msg = _(
                 "This invoice exceeds the financial risk.\n")
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -4,6 +4,7 @@
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ResPartner(models.Model):
@@ -17,10 +18,13 @@ class ResPartner(models.Model):
     risk_invoice_draft_include = fields.Boolean(
         string='Include Draft Invoices', help='Full risk computation')
     risk_invoice_draft_limit = fields.Monetary(
-        string='Limit In Draft Invoices', help='Set 0 if it is not locked')
+        string='Limit In Draft Invoices',
+        currency_field="risk_currency_id",
+        help='Set 0 if it is not locked')
     risk_invoice_draft = fields.Monetary(
         compute='_compute_risk_invoice',
         string='Total Draft Invoices',
+        currency_field="risk_currency_id",
         help='Total amount of invoices in Draft or Pro-forma state')
     risk_invoice_open_include = fields.Boolean(
         string='Include Open Invoices/Principal Balance',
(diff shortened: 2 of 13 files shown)


MiquelRForgeFlow:
cc: @pedrobaeza @JordiBForgeFlow

JordiBForgeFlow:
This is a must feature for us. In our experience companies dealing with international customers can choose tonset credit limits using the currency in which they handle the transactions with that customer.

Perhaps it is still interesting to compute the total risk exposure in company currency so that the company can assess the total risk exposure on an aggregated level. But it's also interesting to assess the aggregated risk exposure per currency

JordiBForgeFlow:
@carlosdauden @pedrobaeza as discussed, this is the proposed addition to the credit risk. For us it is fundamental. And for you it has no further impact, because the default behavior is the same. Please let me know if you are good with this approach. We intend to add more tests and forward-port to 12.0 and 13.0 afterwards.

account_financial_risk/models/res_partner.py
```diff
@@ -155,13 +226,24 @@ def _compute_risk_invoice(self):
...
+                group["commercial_partner_id"][0], self._prefetch)
+            amount = group["amount_total_company_signed"]
+            company_currency = self.env['res.company'].browse(
+                group['company_id'][0]).currency_id
+            if partner.risk_currency_id == company_currency:
+                partner.risk_invoice_draft = amount
+            else:
+                partner.risk_invoice_draft = currency_model._compute(
+                    company_currency, partner.risk_currency_id, amount,
+                    round=False)
```
carlosdauden: Why use _compute private method instead of compute?:
company_currency.compute(amount, partner.risk_currency_id)

account_financial_risk/models/account_invoice.py
```diff
@@ -11,17 +11,28 @@ def risk_exception_msg(self):
...
+            risk_amount = self.amount_total_company_signed
+        else:
+            risk_amount = self.env["res.currency"].with_context(
+                date=self.date_invoice or
+                fields.Date.context_today(self))._compute(
+                self.company_id.currency_id,
+                partner.risk_currency_id,
+                self.amount_total_company_signed,
+                round=False,
+            )
```
carlosdauden: Why use _compute private method instead of compute?

account_financial_risk/models/res_partner.py
```diff
@@ -85,26 +95,86 @@ class ResPartner(models.Model):
...
+        currency_field="risk_currency_id",
+        help='Sum of total risk included')
     risk_exception = fields.Boolean(
         compute='_compute_risk_exception',
         search='_search_risk_exception',
         string='Risk Exception',
-        help='It Indicate if partner risk exceeded')
+        help='It Indicate if partner risk exceeded',
+    )
+    risk_amount_exceeded = fields.Monetary(
```
carlosdauden: Why is this field necessary?
Where is it used?

account_financial_risk/models/res_partner.py
```diff
@@ -226,39 +309,72 @@ def _prepare_risk_account_vals(self, groups):
         # Partner receivable account determines if amount is in invoice field
         for reg in groups['open']['read_group']:
             if reg['partner_id'][0] != self.id:
-                continue
+                continue  # pragma: no cover
+            account = self.env['account.account'].browse(reg['account_id'][0])
             if self.property_account_receivable_id.id == reg['account_id'][0]:
-                vals['risk_invoice_open'] += reg['amount_residual']
+                if self.risk_currency_id == account.company_id.currency_id:
```
carlosdauden: You can use a variable (amount_residual) and move this if 3 lines above to don't repeat the code.
```
if self.risk_currency_id == account.company_id.currency_id:
    amount_residual = reg['amount_residual']
else:
    amount_residual =  account.company_id.currency_id.compute(reg['amount_residual'], self.risk_currency_id)
```
And after only replace reg['amount_residual'] by new amount_residual variable.

account_financial_risk/models/res_partner.py
```diff
@@ -226,39 +309,72 @@ def _prepare_risk_account_vals(self, groups):
...
-            if not partner.customer:
-                partner.risk_exception = False
-                continue
             amount = 0.0
+            amount_exceeded = 0.0
             risk_exception = False
+            if not partner.customer:
+                partner.risk_total = amount
+                partner.risk_amount_exceeded = amount_exceeded
+                partner.risk_exception = risk_exception
```
carlosdauden: I think that this is not needed en versions previous to 13 because if a field not assigned in compute takes the default value.

account_financial_risk/views/res_partner_view.xml
```diff
@@ -20,54 +20,68 @@
                                        attrs="{'readonly': [('risk_allow_edit', '=', False)]}"/>
                                 <button name="open_risk_pivot_info" type="object" class="oe_link"
                                         context="{'open_risk_field': 'risk_invoice_draft'}">
-                                    <field name="risk_invoice_draft" nolabel="1"/>
+                                    <field name="risk_invoice_draft" nolabel="1" widget='monetary' options="{'currency_field': 'risk_currency_id'}"/>
```
carlosdauden, line 23: I think this is not necessary because it is already established in the field definition

carlosdauden (changes requested):

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: I don't have any preference. Why should use public method instead?

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: It's not used here. Here is only computed. One client wants it for their customization. If I don't add it here, then it's awful the compute extension method.

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: right

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: I know, but it's not incorrect. Then, the forward put will be easier.

account_financial_risk/views/res_partner_view.xml
MiquelRForgeFlow: ok for the options, but the widget is good to add, right?

carlosdauden: Ok. Forget my comment.

account_financial_risk/models/res_partner.py
carlosdauden: But you can add this field in the custom.
The compute for this field is as simply as:
risk_amount_exceeded = max(0.0, risk_total - credit_limit)

account_financial_risk/models/res_partner.py
carlosdauden: Ok

account_financial_risk/models/res_partner.py
carlosdauden: More direct, shorter, fewer parameters....
Homogeneity with the existing code:
https://github.com/OCA/credit-control/blob/f0df9f0bb224d7ef0a27781d9bb7b1c7e8090117/sale_financial_risk/models/sale.py#L94

account_financial_risk/models/account_invoice.py
```diff
@@ -11,17 +11,28 @@ def risk_exception_msg(self):
...
+        if self.company_id.currency_id == partner.risk_currency_id:
+            risk_amount = self.amount_total_company_signed
+        else:
+            risk_amount = self.env["res.currency"].with_context(
+                date=self.date_invoice or
+                fields.Date.context_today(self))._compute(
+                self.company_id.currency_id,
+                partner.risk_currency_id,
+                self.amount_total_company_signed,
+                round=False,
```
carlosdauden: Are you sure you don't want to round the amount?

MiquelRForgeFlow: Why would I want to round it?

carlosdauden (approved):

rafaelbn (approved):

account_financial_risk/models/account_invoice.py
```diff
@@ -1,27 +1,52 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import _, api, models
+from odoo import _, api, fields, models
 
 
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
 
+    company_currency_id = fields.Many2one(store=True)
```
pedrobaeza, line 10: I'm worried about the performance of this when upgrading/installing. Please include a migration script and a pre-init hook that pre-compute this field.

pedrobaeza (changes requested):

account_financial_risk/models/account_invoice.py
MiquelRForgeFlow: I put the migration script in v3.0.0, assuming you will do a "ocabot merge major"

pedrobaeza (approved):

pedrobaeza:
Aborting merge. Please include the pre-init in hooks.py and revert the import.

MiquelRForgeFlow:
@pedrobaeza done :)

PR #73, opened 2020-06-19
[FIX] sale_financial_risk: Compute sudo related field commercial_partner_id

carlosdauden:
In multi-company environments it causes an access error raise.

@Tecnativa TT24422


carlosdauden:
This change is needed in previous versions, but not in later versions in which the related fields are computed in sudo mode.

pedrobaeza (approved):

sergio-teruel (approved):

PR #77, opened 2020-06-22
[10.0] Add module account_invoice_overdue_reminder

alexis-via:


patel1008b:
@alexis-via thanks for this module.

Is there any plan for v13 as well?

alexis-via:
@bhaveshselarka I don't have customers using v13, so it's not in my current roadmap, but it seems there is interest for this module in v13, so maybe we could mutualize the effort.

patel1008b:
@alexis-via thanks a lot for the quick response and information, but recently many projects are moving to v13.

so that's the reason behind it. And main thing is about great functionality with simple configurations.

Issue #74, opened 2020-06-20
[12.0] sale.order in done (lock) state


rafamarpe:
Hi,
Is it possible to make this modification?.
I think that modifying these lines could be solved.

- sale_financial_risk/models/sale.py (line 72)
> def _compute_risk_amount(self):
>         for line in self:
>             **if line.state not in ('sale', 'done') and (
>                 line.invoice_status != 'invoiced'):**
>                 line.risk_amount = 0.0
>                 continue

- sale_financial_risk/models/res_partner.py (line 24)
> return self._get_risk_company_domain() + [
>             **('state', 'in', ('sale', 'done'))**,
>             ('commercial_partner_id', 'in', commercial_partners.ids),
>         ]

In this way, a completed order would mean that your invoice status was invoiced

Thanks!

carlosdauden:
This is precisely done so that the user has a mechanism that excludes from the risk orders that have not been fully served and will no longer be served.
The invoiced state cannot be applied if 10 units are ordered, 1 is served and invoice is already in that state.

If you consider it appropriate, you can do a PR by adding a configuration option to determine if orders are included in the done state.

rafamarpe:
If we add an option when automatic order blocking is chosen, which indicates whether to take these orders into account at risk?

carlosdauden:
I prefer that you extract this condition to one function and you add option at company level or system parameter:
https://github.com/OCA/credit-control/blob/12.0/sale_financial_risk/models/sale.py#L73

After this, if you need aditional flow, you can extend that funcion in other module or customer extension.

carlosdauden:
Fixed in https://github.com/OCA/credit-control/pull/78

PR #78, opened 2020-07-08
[12.0][IMP]sale_financial_risk: Add sale orders in done state to risk calculation if sale auto done is enabled.

jesusVMayor:
https://github.com/OCA/credit-control/issues/74

cc: @rafamarpe @carlosdauden

### DIFF
--- a/sale_financial_risk/models/res_config_settings.py
+++ b/sale_financial_risk/models/res_config_settings.py
@@ -1,0 +1,10 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    include_risk_sale_order_done = fields.Boolean(
+        "Include locked sale orders into risk calculation",
+        config_parameter='sale_financial_risk.include_risk_sale_order_done')
--- a/sale_financial_risk/models/sale.py
+++ b/sale_financial_risk/models/sale.py
@@ -37,6 +37,14 @@ class SaleOrder(models.Model):
                 }).action_show()
         return super(SaleOrder, self).action_confirm()
 
+    @api.model
+    def _get_risk_states(self):
+        risk_states = ['sale']
+        if self.env['ir.config_parameter'].sudo().get_param(
+                'sale_financial_risk.include_risk_sale_order_done'):
+            risk_states.append('done')
+        return risk_states
+
 
 class SaleOrderLine(models.Model):
     _inherit = 'sale.order.line'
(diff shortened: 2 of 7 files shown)


sale_financial_risk/models/sale.py
```diff
@@ -37,6 +37,13 @@ def action_confirm(self):
                 }).action_show()
         return super(SaleOrder, self).action_confirm()
 
+    @api.model
+    def _get_risk_states(self):
+        risk_states = ['sale']
+        if self.env['ir.config_parameter'].sudo().get_param('sale.auto_done_setting'):
```
pedrobaeza: System parameters usually use the module name as prefix, so I propose instead: `sale_financial_risk.include_done_orders`. Anyway, not sure if this should be a configuration per company instead of at system level.

jesusVMayor: I use an existing parameter https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L746 i can make an independent configuration, but i  only see the necessity of add the done sale orders to the risk calculation if auto done is set.

pedrobaeza: OK, then I let original author to decide

carlosdauden: Since we modify this we can use a new parameter to give more flexibility to the flows.

gdgellatly: I don't even see the point of this much detail. It should just calculate all sale orders in sale or done state regardless of setting. Whether auto done is enabled or not a user can still lock a sale order without having invoiced it. The 2 states are essentially the same for risk calculation purposes.

carlosdauden: @gdgellatly this is comented here:
https://github.com/OCA/credit-control/issues/74

sale_financial_risk/models/res_config_settings.py
```diff
@@ -0,0 +1,10 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    include_done_orders = fields.Boolean(
```
carlosdauden: The name of this field is too generic.
Note that we are adding a field to the general class res.config.settings, so it would be convenient to refer to risk and sale.
We also try to avoid plurals.
Use include_risk_sale_order_done or something similar.

carlosdauden (approved):

carlosdauden:
:+1:

pedrobaeza:
@gdgellatly what is your final statement on this?

rafamarpe:
Hi,
This programming can be merged?

Thanks

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

please forward port-it to v13

pedrobaeza:
Cherry-pick to v13 at 3bca253f00406bebd6ad1c1419996c47db26a70a and to v14 at f191da8a84ab3f41432909260afceb4979b5492a

PR #86, opened 2020-08-11
[IMP+FIX] account_credit_control: traceable communications

yajo:
Before this patch, `credit.control.communication` was a `TransientModel` that sent emails. This produced that, when someone responded to one of those emails, Odoo detected it was a response but couldn't know where to store it, so it was lost. Also, users with credit control access got a 500 error when clicking on the "View" button in the email.

Now, `credit.control.communication` is a normal `Model` which inherits from `mail.thread` and `mail.activity.mixin`. This makes more sense, because when you are tracking credit control communications, you usually communicate with a customer regarding several lines at once (no sense on asking a customer to pay each line separately).

More changes:

- `credit.control.run` now displays an additional smart button that leads users to generated communications.
- `credit.control.line` now is linked to a communication instead of a simple message (the old link is preserved in a hidden legacy column).
- Some fields are renamed, as they didn't follow guidelines.
- Some menu names are less verbose now.
- Mail templates code is migrated.

@Tecnativa TT24841


account_credit_control/demo/res_users.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Jairo Llopis
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<data noupdate="1">
+
+    <record id="base.user_admin" model="res.users">
```
pedrobaeza: This shouldn't be demo data, but regular XML in `security` folder.

account_credit_control/models/credit_control_communication.py
```diff
@@ -5,59 +5,59 @@
...
     )
-    credit_control_line_ids = fields.Many2many(
+    credit_control_line_ids = fields.One2many(
         comodel_name='credit.control.line',
-        rel='comm_credit_rel',
+        inverse_name="communication_id",
         string='Credit Lines',
     )
-    contact_address = fields.Many2one(
+    contact_address_id = fields.Many2one(
```
pedrobaeza, line 42: There's a chance that the records on the table are not vacuum, so better to rename these fields in scripts for not losing references. Shouldn't this field be required by the way?

account_credit_control/models/credit_control_communication.py
```diff
@@ -85,43 +90,35 @@ def _compute_total(self):
...
-        contact = self.contact_address
+        contact = self.contact_address_id
         email = contact.email
         if not email and contact.commercial_partner_id.email:
             email = contact.commercial_partner_id.email
         return email
 
     @api.multi
     @api.returns('res.partner')
-    def get_contact_address(self):
```
pedrobaeza, line 112: You shouldn't rename this method, as you get a record, not an ID

account_credit_control/models/credit_control_communication.py
```diff
@@ -180,60 +177,23 @@ def _generate_comm_from_credit_lines(self, lines):
...
-                'datas_fname': att[0],
-                'res_model': 'mail.mail',
-                'res_id': email.id,
-                'type': 'binary',
-            }) for att in attachment_list]
-            emails |= email
-        return emails
+            comm.message_post_with_template(
+                comm.policy_level_id.email_template_id.id,
+                composition_mode="mass_post",
```
pedrobaeza, line 194: Is this the suitable mode for having the thread available for answers?

account_credit_control/models/credit_control_run.py
```diff
@@ -86,6 +90,9 @@ def _compute_credit_control_count(self):
                   for data in fetch_data}
         for rec in self:
             rec.credit_control_count = result.get(rec.id, 0)
+            rec.credit_control_communication_count = len(
```
pedrobaeza, line 95: Please don't add this non optimized way. Can you do a `read_group`?

account_credit_control/report/account_credit_control_analysis.xml
```diff
@@ -42,7 +42,7 @@
     </record>
 
     <record model="ir.ui.menu" id="credit_control_analysis_menu">
-        <field name="name">Credit Control Analysis</field>
```
pedrobaeza, line 45: I think this is OK to be named the full way, but change the parent to the `Reports` one.

pedrobaeza (changes requested):
Maybe you can add little checks in tests for count or existence of the records.

pedrobaeza:
I forgot the runbot warning:
```
WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (credit_control_communication_count, credit_control_count) of credit.control.run() have the same label: # of Credit Control Lines.
```

account_credit_control/models/credit_control_communication.py
yajo: > There's a chance that the records on the table are not vacuum, so better to rename these fields in scripts for not losing references

Good catch, but actually in that case, this table would become problematic, because it would contain data that was not meant to be persistent. What I'm going to do is force a vacuum in the pre-migration.

> Shouldn't this field be required by the way?

It's used automatically when the action is executed, and in those cases is always filled automatically. From then, the only thing that actually matters is message followers, so I think it is OK as it is.

account_credit_control/models/credit_control_communication.py
yajo: Oops, this rename came from a bulk search/replace. But I'm seeing that this method does nothing, so I'm just gonna add the needed migration code and remove this method.

account_credit_control/models/credit_control_communication.py
yajo: Yes, it's like posting a normal message, but it uses the mail queue instead of posting directly.

account_credit_control/models/credit_control_run.py
yajo: Not here, because I'd need to count distinct values for the `communication_id` field. The only possibly optimized way to get that AFAIK is raw SQL. Since this is only to be executed in form views, which shouldn't have many values, I didn't see this as a big problem.

Should I do it through SQL?

yajo (commented):
All attended/responded.

sbidoul:
cc/ @qgroulard

qgroulard (approved):
Most welcome improvements, thank you! (code review)

yajo:
New changes:

- Comunication emails have a specific subtype.
- `credit.control.line` has a new state "queued", which indicates that the communication is executed but still not sent. When sent, it's changed to "done" or "email_error".

@rafaelbn review again please.

account_credit_control/data/data.xml
```diff
@@ -4,21 +4,17 @@
...
-        </field>
+        <field name="subject">Credit Control: (${object.policy_level_id.name or 'n/a'})</field>
         <field name="email_to">${object.get_email() or  ''}</field>
         <field name="model_id" ref="model_credit_control_communication"/>
-        <field name="auto_delete" eval="True"/>
-        <field name="lang">
-            ${object.get_contact_address().lang or 'en_US'}
-        </field>
+        <field name="auto_delete" eval="False"/>
+        <field name="lang">${object.contact_address_id.lang or 'en_US'}</field>
```
chienandalu, line 11: This change needs migration script or keeping the original method in the affected model so current templates (they're noupdate=1) don't fail

account_credit_control/models/credit_control_communication.py
```diff
@@ -85,44 +90,29 @@ def _compute_total(self):
...
-        contact = self.contact_address
+        contact = self.contact_address_id
         email = contact.email
         if not email and contact.commercial_partner_id.email:
             email = contact.commercial_partner_id.email
         return email
 
-    @api.multi
-    @api.returns('res.partner')
-    def get_contact_address(self):
```
chienandalu, line 112: If you remove this, you crash current mail templates. We should keep it or do a migration script

pedrobaeza: It can be kept with a TODO for v13/v14

account_credit_control/data/data.xml
```diff
@@ -4,21 +4,17 @@
...
         <field name="report_template" ref="credit_control_summary"/>
         <field name="body_html"><![CDATA[
-  Dear ${object.contact_address.name or ''}
+  Dear ${object.contact_address_id.name or ''}
   <br/>
   <br/>
-  ${object.current_policy_level.custom_mail_text | safe}
+  ${object.policy_level_id.custom_mail_text | safe}
```
chienandalu, line 17: This change breaks no updated templates as well

rafaelbn (changes requested):
After reviewing with @chienandalu this PR need some fixes for who are currently using it as when you send the emails, templates fails. Don't fail in runbot just for existing databases.

rafaelbn:
@Yajo 

1. we need to allow Credit control Manager to change state from "Todo" to "Draft".

If state is todo and user is manager then allow change state to draft



2. Also we should review why the emails to customers are sent as Internal notes

3. Improve test -  codecov/patch — 54.23% of diff hit (target 90.72%) 

In the last test this are the final points.

Thanks

account_credit_control/models/credit_control_run.py
etobella: Well, you could add the one2many to communication and add the credit_control_run_id on communication

etobella (approved):
I think this is a great improvement!!! Can we help with tests or something?

pedrobaeza:
@Yajo please check latest comments and let's merge this.

account_credit_control/data/data.xml
yajo: Yes, this is handled in the post-migration script in this PR, did you see it?

account_credit_control/models/credit_control_communication.py
yajo: Did you see the post-migration script that changes occurrences for this method to the new one?

account_credit_control/data/data.xml
yajo: Hmm same thing.. did you see the migration script?

account_credit_control/models/credit_control_communication.py
pedrobaeza: The problem is not this module, but others relying on this one. Please keep the method with the TODO.

yajo:
> we need to allow Credit control Manager to change state from "Todo" to "Draft".

Fixed with last commit.

You just have to go to *Action > Change Lines' State*

yajo:
>     1. Also we should review why the emails to customers are sent as Internal notes

It seems [the message subtype must be `mail.mt_comment` if we want it as a normal discussion comment](https://github.com/odoo/odoo/blob/f0508b3fd43339241c3a75273871ddc06690db31/addons/mail/models/res_partner.py#L72).

That's not very nice because it means that I couldn't use the `account_credit_control.mt_request` mail subtype, which adds some features such as detecting if a queued message was an actual request or not, to set the process from queued to done.

However, the fact it appears as grayed-out doesn't mean it's really a note. Maybe it's just Odoo distinguishing between messages written by users from messages written by the system, isn't it?

pedrobaeza:
It doesn't sound to me that Odoo grays out that. Anyway, it should be non gray for proper UX. Can you please investigate the reason? Also the subtype can be important to be respected.

etobella:
The gray is defined on:

https://github.com/odoo/odoo/blob/f0508b3fd43339241c3a75273871ddc06690db31/addons/mail/models/mail_message.py#L526

The white is set if is_discussion or is_notification are true.

So, at the end, the subtype needs to be comment (as is_discussion will not be an option, as there is model and resource) or the function must be inherited and modifed :thinking:

yajo:
Thanks @etobella, it confirms what I said in https://github.com/OCA/credit-control/pull/86#issuecomment-689408353.

So this leaves us 2 options:

1. Modify the mail stack to display some subtypes as messages.
2. Leave it as it is and assume this is how Odoo works.

I think the correct option is 2. At least for this PR.

pedrobaeza:
But can we put any subtype that puts things as not grey as Odoo does with mail templates manually sent for example?

etobella:
Not easily :thinking: I added on MailMessage the following code and worked as expected:
``` python
    @api.multi
    def message_format(self):
        result = super().message_format()
        credit_control = self.env['ir.model.data'].xmlid_to_res_id(
            'account_credit_control.mt_request'
        )
        for message in result:
            message.update({
                'is_discussion': message['is_discussion'] or (
                    message['subtype_id'] and
                    message['subtype_id'][0] == credit_control
                )
            })
        return result
```

etobella:
If you want I can make a PR solving this

Also, I found a small issue regarding multicompany behavior :thinking:

etobella:
I made the PR :wink:

pedrobaeza (approved):

pedrobaeza:
/ocabot merge major

@Yajo now you need to migrate this to v13 :scream:

PR #83, opened 2020-07-31
[12.0][IMP] Add selection: choose which currency should be used in risk

MiquelRForgeFlow:
Forward port of https://github.com/OCA/credit-control/pull/70.

PR #89, opened 2020-09-03
[13.0][IMP] Add selection: choose which currency should be used in risk

MiquelRForgeFlow:
Forward-port of https://github.com/OCA/credit-control/pull/83.

Please, use **merge major**!

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -4,6 +4,7 @@
 from dateutil.relativedelta import relativedelta
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ResPartner(models.Model):
@@ -18,11 +19,14 @@ class ResPartner(models.Model):
         string="Include Draft Invoices", help="Full risk computation"
     )
     risk_invoice_draft_limit = fields.Monetary(
-        string="Limit In Draft Invoices", help="Set 0 if it is not locked"
+        string="Limit In Draft Invoices",
+        currency_field="risk_currency_id",
+        help="Set 0 if it is not locked",
     )
     risk_invoice_draft = fields.Monetary(
         compute="_compute_risk_account_amount",
         string="Total Draft Invoices",
+        currency_field="risk_currency_id",
         help="Total amount of invoices in Draft or Pro-forma state",
     )
     risk_invoice_open_include = fields.Boolean(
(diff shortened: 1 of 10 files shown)


AaronHForgeFlow (approved):
:+1: Functional review works as expected.

pedrobaeza:
Trying on runbot, I see something weird:

Peek 04-09-2020 08-59

Changing the currency in the contact, I would expect the amount of opened operations to change according that currency.

MiquelRForgeFlow:
@pedrobaeza that's not weird. You are editing and playing with onchange (NewIds). Once you save, then the values will be updated accordingly.

pedrobaeza:
OK, but I think this can be improved adding the field to the depends list for getting automatic update even on edit mode. Is that possible or there are `inverse` methods involved?

MiquelRForgeFlow:
@pedrobaeza your wish came true  🧞

account_financial_risk/models/res_partner.py
```diff
@@ -108,7 +123,71 @@ class ResPartner(models.Model):
...
+        for partner in self:
+            if partner.credit_currency == "manual":
+                partner.manual_credit_currency_id = partner.risk_currency_id
+            else:
+                partner.manual_credit_currency_id = False
+
+    @api.onchange("risk_currency_id")
+    def _onchange_risk_currency_id(self):
+        for partner in self:
+            partner._compute_risk_account_amount()
```
pedrobaeza: Can you do

```suggestion
        self._compute_risk_account_amount()
```
?

sale_financial_risk/models/res_partner.py
```diff
@@ -37,14 +40,30 @@ def _compute_risk_sale_order(self):
...
+                company,
+                fields.Date.context_today(self),
+                round=False,
+            )
+
+    @api.onchange("risk_currency_id")
+    def _onchange_risk_currency_id(self):
+        super()._onchange_risk_currency_id()
+        for partner in self:
+            partner._compute_risk_sale_order()
```
pedrobaeza:
```suggestion
        self._compute_risk_sale_order()
```
?

account_financial_risk/models/res_partner.py
MiquelRForgeFlow: done

pedrobaeza (approved):

pedrobaeza:
FYI, @MiquelRForgeFlow @JordiBForgeFlow making the field `company_currency_id` stored has provoked an error very difficult to debug, but finally found and fixed removing such thing in direct commit: https://github.com/OCA/credit-control/commit/216baf808587ab08b8f02e0463a8f19e99d0b2c5

I see no real reason for such store, but if so, it should be done on other field, not reusing standard v13 one. Maybe it has been a chain of coincidences, as in v12, there wasn't such dynamic lines recomputation that needs to have immediately the field populated, and maybe is a fault of the framework not populating it on time, but the question is that it breaks standard flows. Please check why you did that way and let's discuss about it.

pedrobaeza:
Cherry-picked for v14 in 94d9ad9a7b21c5fba63af7cbc7978e16be26cb65 and v15 in 784842163ba3db1ff68bcf166da1dd6505a59d03

PR #90, opened 2020-09-03
[12.0][FIX] hook: check if column exists

MiquelRForgeFlow:
Please, do "**merge major**" because last merge in https://github.com/OCA/credit-control/pull/83 was done with merge patch.


pedrobaeza:
For lower diff, do if ...: return and don't change indentation.

MiquelRForgeFlow:
Ready

Issue #84, opened 2020-08-04
sale_financial_risk: partner_invoice_id issue


abirbasolt:
The partner_Invoice_id is who is going to pay, so would be more logical to check the credit from the Partner_invoice_id when you confirm the order.

pedrobaeza:
I told you about opening 2 separate issues. Let's keep this one for the first, so I edited the original text.

pedrobaeza:
@carlosdauden what do you think about this? I think it can be logical to put the risk on the invoicing partner, not the ordering one.

carlosdauden:
I agree

pedrobaeza:
@abirbasolt you can open then a PR solving the problem.

carlosdauden:
Fixed in https://github.com/OCA/credit-control/pull/92

pedrobaeza:
Solved in 13.0, so closing

PR #92, opened 2020-09-22
[13.0][FIX+IMP] sale_financial_risk: Compute risk for partner invoice. Inheritable method.

carlosdauden:
This PR solves https://github.com/OCA/credit-control/issues/84

@Tecnativa TT25596

### DIFF
--- a/sale_financial_risk/models/sale.py
+++ b/sale_financial_risk/models/sale.py
@@ -8,29 +8,33 @@ from odoo.tools import float_round
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    def evaluate_risk_message(self, partner):
+        risk_amount = self.currency_id._convert(
+            self.amount_total,
+            self.company_id.currency_id,
+            self.company_id,
+            self.date_order
+            and self.date_order.date()
+            or fields.Date.context_today(self),
+            round=False,
+        )
+        exception_msg = ""
+        if partner.risk_exception:
+            exception_msg = _("Financial risk exceeded.\n")
+        elif partner.risk_sale_order_limit and (
+            (partner.risk_sale_order + risk_amount) > partner.risk_sale_order_limit
+        ):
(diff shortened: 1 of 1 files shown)


victoralmau (approved):

marcelsavegnago (approved):
LGTM

ernestotejeda (approved):

sale_financial_risk/models/sale.py
```diff
@@ -8,29 +8,34 @@
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    def evaluate_risk_message(self):
+        partner = self.partner_invoice_id.commercial_partner_id
```
pedrobaeza: As you are extracting that code, decouple it passing the partner as argument:
```suggestion
    def evaluate_risk_message(self, partner):
```

sale_financial_risk/models/sale.py
```diff
@@ -8,29 +8,34 @@
...
-            elif partner.risk_sale_order_limit and (
-                (partner.risk_sale_order + risk_amount) > partner.risk_sale_order_limit
-            ):
-                exception_msg = _("This sale order exceeds the sales orders risk.\n")
-            elif partner.risk_sale_order_include and (
-                (partner.risk_total + risk_amount) > partner.credit_limit
-            ):
-                exception_msg = _("This sale order exceeds the financial risk.\n")
+            partner = self.partner_invoice_id.commercial_partner_id
+            exception_msg = self.evaluate_risk_message()
```
pedrobaeza: Following the other change:
```suggestion
            exception_msg = self.evaluate_risk_message(partner)
```

pedrobaeza (changes requested):

pedrobaeza (approved):

sale_financial_risk/models/sale.py
carlosdauden: Ups, that was the idea...

PR #94, opened 2020-10-02
[IMP] account_credit_control: Set multicompany usage. Remove warnings

etobella:
I have detected that the generation was not working properly on multicompany environments, this PR tries to solve the issue

### DIFF
--- a/account_credit_control/models/credit_control_policy.py
+++ b/account_credit_control/models/credit_control_policy.py
@@ -67,23 +67,27 @@ class CreditControlPolicy(models.Model):
             record.policy_sign = mapping[policy_sign]
 
     @api.multi
-    def _move_lines_domain(self, controlling_date):
+    def _move_lines_domain(self, credit_control_run):
         """ Build the default domain for searching move lines """
         self.ensure_one()
         domain = [
             ('account_id', 'in', self.account_ids.ids),
             ('reconciled', '=', False),
             ('partner_id', '!=', False),
+            ('company_id', '=', credit_control_run.company_id.id),
         ]
+        # We need to set the company in order to work properly with multi-companies.
+        # If we have Company A and Company B (child of A), we might be able to run this
+        # for company B from Company A view.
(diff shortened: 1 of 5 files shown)


jarroyomorales (approved):
Code review LGTM! :+1:

account_credit_control/models/credit_control_policy.py
```diff
@@ -67,23 +67,24 @@ def _compute_policy_sign(self):
...
     @api.multi
-    def _move_lines_domain(self, controlling_date):
+    def _move_lines_domain(self, credit_control_run):
         """ Build the default domain for searching move lines """
         self.ensure_one()
         domain = [
             ('account_id', 'in', self.account_ids.ids),
             ('reconciled', '=', False),
             ('partner_id', '!=', False),
+            ('company_id', 'child_of', credit_control_run.company_id.id),
```
pedrobaeza: Why is this needed? This is controlled by record rules, and this is not run in sudo mode.

etobella: Let's assume you have Company A and its Child Company B and you are working on company A. By rules, you can create a Credit Control Run for company B, but when executed, It generates the moves for A and B. This solves this kind of issue

pedrobaeza: But why `child_of`? If the credit control run is for A, it will add moves for A and B.

etobella: True, I assumed it was the right option, but we can change it if you think it is better to make the company forced (both options are right for me)

pedrobaeza: I would say to do it strict, and please put a comment depicting the use case you have mentioned for avoiding later doubts.

etobella: Done!

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Please FW port it to v13

PR #97, opened 2020-10-15
[13.0][IMP] sale_financial_risk: Change related risk partner target and rename field

carlosdauden:
This change allow compute risk for partner invoice to finish https://github.com/OCA/credit-control/pull/92

@Tecnativa


pedrobaeza (changes requested):
For performance reasons on existing installations, you need to add a pre-migration script that call the same hook than on post-install.

MiquelRForgeFlow (approved):

pedrobaeza:
@carlosdauden plase make the change on performance part and we can merge this.

carlosdauden:
@pedrobaeza Changes done

pedrobaeza (approved):

PR #106, opened 2020-10-29
[MIG] sale_financial_risk: Migration to 14.0

psugne:
@Tecnativa


alan196 (approved):
Tested functionally in my local :+1:

alan196 (changes requested):
Please consider change from PR https://github.com/OCA/credit-control/pull/111

alan196 (approved):
👍

pedrobaeza:
Why have you removed the pre-init hook?

psugne:
@pedrobaeza Seems pre_init_hook was used in [13.0.2.0.2 pre-migration](https://github.com/versada/credit-control/blob/13.0/sale_financial_risk/13.0.2.0.2/pre-migration.py) script and as pre_init_hook by itself for v13.0 because of newly created [risk_partner_id](https://github.com/versada/credit-control/blob/13.0/sale_financial_risk/models/sale.py#L70-L75) field to be sure that field will be added. Please correct me if I'm wrong.

pedrobaeza:
It was both used as pre-init hook and for the migration script. The goal of it is to speed-up the computation of that field that was added in such version, so when migrating from the previous version to that one, we call the same method for not having slow update times. Thus, it's still needed in v14, although the migration scripts part is not needed anymore. Please restore it.

psugne:
> It was both used as pre-init hook and for the migration script. The goal of it is to speed-up the computation of that field that was added in such version, so when migrating from the previous version to that one, we call the same method for not having slow update times. Thus, it's still needed in v14, although the migration scripts part is not needed anymore. Please restore it.

Thanks for explanation. I have just restored it.

pedrobaeza (approved):

PR #107, opened 2020-10-29
[MIG] stock_financial_risk: Migration to 14.0

psugne:
@Tecnativa


alan196 (approved):
Tested functionally in my local 👍

psugne:
Could you check please @pedrobaeza

pedrobaeza:
Several things:

- Rebase for getting all the dependencies.
- The commit order is not correct, as you have to follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 procedure, and let all the pre-commit stuff (if any), before. I'm seeing a commit at the end called `black` that I don't know if it's because of your changes or due to the change of black version in v14.

pedrobaeza:
Please squash this together:

Selección_012

psugne:
> Several things:
> 
> * Rebase for getting all the dependencies.
> * The commit order is not correct, as you have to follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 procedure, and let all the pre-commit stuff (if any), before. I'm seeing a commit at the end called `black` that I don't know if it's because of your changes or due to the change of black version in v14.

@pedrobaeza thank you for turbo answer! Rebased from 14.0 and squashed commits.

stock_financial_risk/models/stock.py
```diff
@@ -17,10 +17,12 @@ def _action_done(self, cancel_backorder=False):
             )
             if moves:
                 raise exceptions.UserError(
-                    _("Financial risk exceeded in partner:\n%s")
-                    % moves.mapped("partner_id.name")
+                    _(
+                        "Financial risk exceeded in partner:\n%s",
+                        moves.mapped("partner_id.name"),
```
pedrobaeza, line 22: This have to be outside of `_()`, or no translation will be done.

pedrobaeza (changes requested):

stock_financial_risk/models/stock.py
psugne: @pedrobaeza could you recheck please. IIUC v14.0 [supports](https://www.odoo.com/documentation/14.0/reference/translations.html#variables) passing arguments to translation lookup.

pedrobaeza: OK, I see, thanks for the pointer.

pedrobaeza (approved):

pedrobaeza:
@sbidoul the template must be updated for proper pre-commit?

sbidoul:
@pedrobaeza here that is something new:

```
    Collecting docutils<0.15.1.1,>=0.15.1

      Downloading docutils-0.15.1-post1.tar.gz (1.7 MB)

    

stderr:

    ERROR: Requested docutils<0.15.1.1,>=0.15.1 from https://files.pythonhosted.org/packages/d4/12/6c3fd74a590c7327c98cae008c11d536029fa9cd7924de477e8cb8804186/docutils-0.15.1-post1.tar.gz#sha256=f33ddb723332c6d6b6d99731ee1fc0c35eb4044a2df5cca1c64c8aa78eaf22cb (from oca-maintainers-tools==0.1.dev998+g1b5c7ad) has different version in metadata: '0.15.1'
```

This is due to the latest pip resolver being stricter and erroring on docutils which announces version 0.15.1-post1 on PyPI but actually has version 0.15.1 in it's metadata. Annoying.

pedrobaeza:
Uf, then do you figure out any solution in our hands?

sbidoul:
We are back to https://github.com/OCA/maintainer-tools/issues/423 I guess...

pedrobaeza:
Isn't already pinned?

sbidoul:
Yes but if we need to change the docutils version to work around the new pip restriction that issue will reappear. I'll first discuss it with the pip team. https://github.com/pypa/pip/issues/9203#issuecomment-740718379

PR #110, opened 2020-11-04
[13.0][FIX] account_credit_control avoid cancelled moves

CasVissers-360ERP:
A credit control run can currently contain != posted moves (i.e. cancelled moves). This fix proposes adding a domain to fix this.

@pedrobaeza


pedrobaeza (approved):

MiquelRForgeFlow (approved):
Not just it's excluding cancelled moves, also it's excluding draft moves.

PR #111, opened 2020-11-19
[FIX] sale_financial_risk: fixed bug when the line of sale order is a…

jesus01x:
Step to reproduce

add lines to a sale order than be a note or a section. 

current behavior

When try to confirm the sale order get this error : 
"AssertionError: precision_rounding must be positive, got 0.0"

expected behavior

Can confirm the sale order without problems.


alan196 (approved):
Functional and technical review 👍 I confirm the issue

pedrobaeza (approved):

PR #122, opened 2021-02-25
[13.0][IMP] account_credit_control - Consider ToDo lines as left lines

CasVissers-360ERP:
When lines are left as ToDo on the next run new lines for the next level are created. All lines draft or ToDo lines should be processed first before creating a new line.

### DIFF
--- a/account_credit_control/models/credit_control_line.py
+++ b/account_credit_control/models/credit_control_line.py
@@ -241,14 +241,15 @@ class CreditControlLine(models.Model):
             )
             lines_to_create.append(vals)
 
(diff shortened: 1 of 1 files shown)


CasVissers-360ERP:
@pedrobaeza 
Wonder how you feel about this change?

pedrobaeza:
My colleague @carlosdauden has to say.

carlosdauden (approved):

carlosdauden:
I had thought about the possibility of adding a configuration option for this behavior, but the truth is that it makes sense and if nobody is bothered we do not complicate the configuration of the module

PR #123, opened 2021-03-02
[IMP] account_financial_risk: dynamically reuse the groups defined on…

wtaferner:
… the view

To be more flexible without the need to inherit or override the computation is checking if the user has one of the view groups in order to skip readonly.
It is the logical step as without seeing the view it would not be possible to edit anyway.

Info: @wt-io-it

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -189,13 +189,15 @@ class ResPartner(models.Model):
                 raise ValidationError(_("Choose Manual Credit Currency."))
 
     def _compute_risk_allow_edit(self):
-        self.update(
-            {
-                "risk_allow_edit": self.env.user.has_group(
-                    "account.group_account_manager"
-                )
-            }
-        )
+        view_id = self.env.ref('account_financial_risk.res_partner_view_risk', raise_if_not_found=False)
+        if view_id:
+            risk_allow_edit = any([
+                self.env.user.has_group(ext_id)
+                for id,ext_id in view_id.groups_id.get_external_id().items()
+            ]) if view_id.groups_id else True
+        else:
+            risk_allow_edit = self.env.user.has_group("account.group_account_manager")
+        self.risk_allow_edit = risk_allow_edit
 
     @api.model
     def _get_risk_company_domain(self):


wtaferner:
@pedrobaeza 
Any idea why Travis failed? Was it because of the failed linting check or some CI issue? Afaics the linter would have done the job for me and continues, right?

pedrobaeza:
It seems you still don't know pre-commit and black auto-formatter that we apply in OCA since 13.0. Please install it locally (https://pre-commit.com/#install) and run `pre-commit run -a` before committing, and the magic will happen.

pedrobaeza:
Why being so reticent to override the method? I see this a bit convoluted sincerely, but I let my colleague @carlosdauden to give his opinion about this.

wtaferner:
@pedrobaeza 
Ok, so it was the linter, thx for the info.

Well, basically I did not get why the readonly exists at all as the view has the group set and nobody else would see the fields by default and as a customer asked me that sales should set these values I failed the easy way (just put the group on the inherited view) and thought it would be a good suggestion, but of course I do not know if there was any deeper meaning behind the hard coded group...really something I suggested...no expectations to be merged...just learning the reasons behind if there were any ;-)

pedrobaeza:
All of that was done by my colleague @carlosdauden and I don't remember well why, but there's a reason :) I let him to explain.

wtaferner:
Damn, and I thought it might be a good suggestion 😄 

But I am really curious as the readonly on the view would only make sense to provide the fields to other groups in case added, but then it is a pity as you could still write any other way as it is only enforced on the view itself...

carlosdauden:
I had the idea of adding a specific group for this.
If you can wait a few weeks, I will do it, but if you want to make the changes to add the group, it will seem fantastic.
You can think of a case that cannot be covered with the specific permission to overcome risk?

wtaferner:
@carlosdauden 
Hmm, to introduce an own group might be a possibility too, but then I assume my solution would kill it as it is generic and you can dynamically add any group (without introducing a new group) you want of the existing groups (like in my case add Sales / All Documents) to the permissions of the view and the group is properly taken into consideration (showing the tab + having the computed field automatically true)

So, I would additionally only check permissions on write to be really complete about the permissions.

What do you think? But yeah, besides, an own group is very fine grained, so I guess we can do both 😄

wtaferner:
@carlosdauden 
Would that work for you that my suggestion is merged and you add an explicit group like you want to have later?
If yes, I would proceed and fix the PR that it is green to merge 😄

PR #126, opened 2021-03-17
[FIX] account_credit_control: do not write on action when mailing

yajo:
Without this patch, the actual action record was overwritten each time communications were mailed, resulting in the "communications" menu being filtered always for last mailings, effectively hiding most of the communications.

@Tecnativa TT28800 TT28823 @chienandalu

### DIFF
--- a/account_credit_control/__manifest__.py
+++ b/account_credit_control/__manifest__.py
@@ -4,7 +4,7 @@
 # Copyright 2019-2020 Tecnativa
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {'name': 'Account Credit Control',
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
Tricky one :sweat_smile:

carlosdauden (approved):

sergio-teruel (approved):

marcelsavegnago (approved):

pedrobaeza (changes requested):
Please add a migration script that resets the domain on the referenced action.

yajo:
Done.

pedrobaeza (approved):

PR #127, opened 2021-03-18
Break with non present number field on invoice

Martronic-SA:
There is no more "number" field is v13. replace it by name instead.
or if its provided by another module, please depend on it.


pedrobaeza (approved):
Yes, this is a bad migration

PR #132, opened 2021-07-13
[13.0][FIX TRANSLATION] - account_credit_control - date in english should be date in french

baimont:


pedrobaeza:
Please use weblate for this:

https://odoo-community.org/page/translate

PR #138, opened 2021-07-23
[13.0][IMP] account_financial_risk: Change risk tab group to new for more flexibility

carlosdauden:
@Tecnativa TT26717

### DIFF
--- a/account_financial_risk/views/res_partner_view.xml
+++ b/account_financial_risk/views/res_partner_view.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<!-- Copyright 2016-2018 Tecnativa - Carlos Dauden
+<!-- Copyright 2016-2021 Tecnativa - Carlos Dauden
      License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
 <odoo>
     <record id="res_partner_view_risk" model="ir.ui.view">
@@ -7,7 +7,10 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="priority" eval="20" />
-        <field name="groups_id" eval="[(4, ref('account.group_account_manager'))]" />
+        <field
+            name="groups_id"
+            eval="[(4, ref('account_financial_risk.group_account_financial_risk_user'))]"
+        />
         <field name="arch" type="xml">
             <page name="sales_purchases" position="after">
                 <page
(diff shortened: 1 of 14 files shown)


account_financial_risk/views/res_partner_view.xml
```diff
@@ -7,7 +7,7 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="priority" eval="20" />
-        <field name="groups_id" eval="[(4, ref('account.group_account_manager'))]" />
+        <field name="groups_id" eval="[(4, ref('account_financial_risk.group_overpass_partner_risk_exception'))]" />
```
pedrobaeza: But non allowed to overpass the risk should query the info (but not change it). I think it should be a new group, and stacked them as User/Manager or similar.

carlosdauden: Before this change only `group_account_manager` can query the info.

After this change, `group_account_manager` can continue to query the info because `group_account_manager` has implicit `group_overpass_partner_risk_exception`:
https://github.com/OCA/credit-control/blob/858d68d3f872144d0d6f644c242131bf0400390d/account_financial_risk/security/security.xml#L8-L13

In all this time, it has not been necessary for someone without group_account_manager role can query the details of the risk.
A new group with permissions to query can cause a change in behavior in a stable version, it can be proposed for a new version if it is considered necessary.

pedrobaeza: It's not a change in behavior with proper migration scripts including account managers in such group. I think it's logical to have a risk viewer and risk manager, don't you think?

Anyways, I'm not blocking.

carlosdauden:
Changes done

carlosdauden:
ping @AaronHForgeFlow @psugne @wtaferner @alan196

pedrobaeza (changes requested):
A post-migration script is missing that adds the financial risk manager to `group_account_manager` users for keeping old behavior:
```python
env["account.group_account_manager"].users.write({"groups_id": [(4, env.ref("account_financial_risk.group_account_financial_risk_manager"))]})
```

Change also the README to specify in CONFIGURE.rst the new group.

carlosdauden:
The behavior is not changed for those who already have the module installed:
(Code before this PR, in this PR is removed, but implied_ids is not affected by this remove)
https://github.com/OCA/credit-control/blob/1e97ae5ac57a0c184ddb775893b22f3bfbf4e30d/account_financial_risk/security/security.xml#L8-L13

For the change you comment on to make sense, it should be completed with the removal of the group in implied_ids.
And doing that does change the behavior when adding an account manager.

pedrobaeza (approved):
OK, sorry for the noise

carlosdauden:
README and CONFIGURE files changed

sergio-teruel (approved):

pedrobaeza:
/ocabot merge nobump

Can you please fw-port it to v14?

PR #143, opened 2021-08-05
Init copier template

joao-p-marques:
Also, add the `swig` package, now needed when installing from `reporting-engine`


victoralmau (approved):

yajo (approved):

pedrobaeza:
Please don't add the burden of swig in all repositories.

PR #148, opened 2021-09-03
[13.0][FIX] account_financial_risk: Show ValidationError when post multiple invoices…

### LINKED ISSUE #147: account_financial_risk: "Expected singleton: res.company(1, 2)" while evaluating 'model._autopost_draft_entries()
Odoo Version:

- [x]  13.0 

I think it can be reproduced in 13.0 and above versions

Hi, the cron that calls the method _autopost_daft_entries() is raising an error due to account_financial_risk.

The error occurs when:

- More than one company in odoo
- More than one draft move in each company with auto_post = True.

The error is:

```
   File "/opt/odoo/auto/addons/account_financial_risk/models/account_invoice.py", **line 65, in post
     or self.company_id.allow_overrisk_invoice_validation**
   File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1021, in __get__
     record.ensure_one()
   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5104, in ensure_one
     raise ValueError("Expected singleton: %s" % self)
 ValueError: <class 'ValueError'>: "Expected singleton: res.company(1, 2)" while evaluating
 'model._autopost_draft_entries()'
```

We fixed the issue (was urgently in our instance ) with:

Original code:

```
    def action_post(self):
        if (
            self.env.context.get("bypass_risk", False)
            or self.company_id.allow_overrisk_invoice_validation
        )
```

Fixed code:

```
    def action_post(self):
        for sel in self:
            if (
                self.env.context.get("bypass_risk", False)
                or sel.company_id.allow_overrisk_invoice_validation
            ):
                return super().post()
```

Don't know if it's the best way to fix the issue.

carlosdauden:
… from wizard. Avoid singleton with multi-company _autopost_draft_entries.
 When we post from validate.account.move wizard, validate_move returns 'ir.actions.act_window_close'.
 Fixed issue #147 

@Tecnativa TT30969

### DIFF
--- a/account_financial_risk/tests/test_account_financial_risk.py
+++ b/account_financial_risk/tests/test_account_financial_risk.py
@@ -4,6 +4,7 @@
 from dateutil.relativedelta import relativedelta
 
 from odoo import fields
+from odoo.exceptions import ValidationError
 from odoo.tests.common import SavepointCase
 
 
@@ -121,7 +122,8 @@ class TestPartnerFinancialRisk(SavepointCase):
         invoice2 = self.invoice.copy({"partner_id": self.invoice_address.id})
         self.assertAlmostEqual(self.partner.risk_invoice_draft, 550.0)
         self.assertAlmostEqual(self.partner.risk_invoice_unpaid, 550.0)
-        wiz_dic = invoice2.post()
+        # Force active_model to simulate action in form view
+        wiz_dic = invoice2.with_context(active_model="account.move").post()
         wiz = self.env[wiz_dic["res_model"]].browse(wiz_dic["res_id"])
         self.assertEqual(wiz.exception_msg, "Financial risk exceeded.\n")
         self.partner.risk_invoice_unpaid_limit = 0.0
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Can you please check Travis?

pedrobaeza (approved):

account_financial_risk/tests/test_account_financial_risk.py
```diff
@@ -121,7 +123,8 @@ def test_invoices(self):
         invoice2 = self.invoice.copy({"partner_id": self.invoice_address.id})
         self.assertAlmostEqual(self.partner.risk_invoice_draft, 550.0)
         self.assertAlmostEqual(self.partner.risk_invoice_unpaid, 550.0)
-        wiz_dic = invoice2.post()
+        invoice_form = Form(invoice2)
+        wiz_dic = invoice_form.post()
```
pedrobaeza: This is a `Form` and can't run methods. I think the problem is that you are posting twice, so the solution is:

```suggestion
```

pedrobaeza (changes requested):

carlosdauden:
Tests fixed

pedrobaeza (approved):

PR #152, opened 2021-09-28
[13.0][FIX] account_financial_risk: Active model not in context from header buttons

carlosdauden:
@Tecnativa TT32194

### DIFF
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -71,8 +71,8 @@ class AccountMove(models.Model):
             if exception_msg:
                 # When we post from validate.account.move wizard, validate_move returns
                 # 'ir.actions.act_window_close'
-                # Check active model to raise exception instead of returns wizard
-                if self.env.context.get("active_model") == "account.move":
+                # Active active_model is false in header buttons as Post in invoice form view
+                if not self.env.context.get("active_model", False):
                     return (
                         self.env["partner.risk.exceeded.wiz"]
                         .create(
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -69,9 +69,16 @@ class AccountMove(models.Model):
         ):
             exception_msg = invoice.risk_exception_msg()
             if exception_msg:
-                # When we post from validate.account.move wizard, validate_move returns
-                # 'ir.actions.act_window_close'
-                # Active active_model is false in header buttons as Post in invoice form view
+                # Active active_model is False if we click in form view header buttons
+                # as 'Post'.
+                # Active model is "account.move" if we select moves from tree view and
+                # we use validate.account.move wizard.
+                # We can not return "partner.risk.exceeded.wiz" because validate_move
+                # don't save post method return and it always returns
+                # "{'type': 'ir.actions.act_window_close'}"
+                # Therefore the only way to show the notice is with raise and to be able
+                # to validate invoices with exceeded risk it must be done from the
+                # form view
                 if not self.env.context.get("active_model", False):
                     return (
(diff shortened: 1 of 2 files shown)


account_financial_risk/models/account_invoice.py
```diff
@@ -71,8 +71,8 @@ def post(self):
             if exception_msg:
                 # When we post from validate.account.move wizard, validate_move returns
                 # 'ir.actions.act_window_close'
-                # Check active model to raise exception instead of returns wizard
-                if self.env.context.get("active_model") == "account.move":
+                # Active active_model is false in header buttons as Post in invoice form view
+                if not self.env.context.get("active_model", False):
```
pedrobaeza, line 82: I think it should be this for covering the previous case:

```suggestion
                if self.env.context.get("active_model", "account.move") == "account.move":
```

carlosdauden: When `post` method is called from validate wizard active_model is "account.move", because wizard is launched from action.

https://github.com/odoo/odoo/blob/26217a1863d08e5aae967246da558aee8b0a32f3/addons/account/wizard/account_validate_account_move.py#L20
https://github.com/odoo/odoo/blob/26217a1863d08e5aae967246da558aee8b0a32f3/addons/account/wizard/account_validate_account_move.py#L10

pedrobaeza: Then I don't get which cases should go by one path or another. I thought you wanted the "interactive" popup for that case and the manual button.

carlosdauden:
ping @chienandalu  @MiquelRForgeFlow

chienandalu (approved):

MiquelRForgeFlow (approved):
Ok, no problem with this approach, but why the old approach was done that way in the first place?

account_financial_risk/models/account_invoice.py
pedrobaeza: Any answer to this?

chienandalu (approved):

carlosdauden:
@pedrobaeza @MiquelRForgeFlow 
Changes done to get more understandable comments.

pedrobaeza:
Can this be cherry-picked as is to v14/v15?

carlosdauden:
The same strategy is needed in v14 and v15:
https://github.com/odoo/odoo/blob/54e54466f0858d4b19fa8f61ff2b927e00f8f5a8/addons/account/wizard/account_validate_account_move.py#L22
https://github.com/odoo/odoo/blob/c6b6dd5650fd3a4b3e6942505075803eaa30a240/addons/account/wizard/account_validate_account_move.py#L22

But mehod name has changed and cherry-pick will collide with this commit:
https://github.com/OCA/credit-control/commit/2438e66117968102d8170ca2303bc27239753f0f

I go to create new PR for v14

pedrobaeza:
Thanks, waiting for such PR.

carlosdauden:
Already forward ported here:
https://github.com/OCA/credit-control/pull/161

pedrobaeza:
Great, and for v15? Was it forward-ported in your last final form?

carlosdauden:
Second commit that only improves the comment can be cherry-picked:
https://github.com/OCA/credit-control/pull/152/commits/c28bee61688ea753d71256767b9c1a67f9cab3a5

MiquelRForgeFlow:
> Already forward ported here: #161

Wrong. Check again haha

MiquelRForgeFlow:
In v13 has: `if not self.env.context.get("active_model", False):`

In v14 has:  `if self.env.context.get("active_model") == "account.move":`

carlosdauden:
Oops, that's right, sorry.
Now I do the PR

PR #156, opened 2021-11-08
Account_financial_risk: FIX, migration to v.15.0

JoanMForgeFlow:


account_financial_risk/tests/test_account_financial_risk.py
```diff
@@ -6,8 +6,10 @@
 from odoo import fields
 from odoo.tests.common import SavepointCase
```
pedrobaeza, line 7: Remove this one

PR #160, opened 2021-11-23
[14.0][IMP] account_financial_risk: consider multicurrency

AaronHForgeFlow:
And some fix.

### DIFF
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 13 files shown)


JoanMForgeFlow (approved):
LG!

account_financial_risk/models/account_invoice.py
```diff
@@ -82,4 +83,43 @@ def action_post(self):
                     )
                     .action_show()
                 )
-        return super().action_post()
+        return super().post()
```
carlosdauden: ¿Why this change?
¿`def action_post(self):` and `return super().post()`?

Please, include this changes:https://github.com/OCA/credit-control/commit/4bd8fcd1f1cdc8120b354f0054782ecb24659948

carlosdauden (changes requested):

account_financial_risk/models/account_invoice.py
carlosdauden: This wizard call to method `_post`:
https://github.com/odoo/odoo/blob/37481a9f797e6dc519405aef1e6196a9e76a9203/addons/account/wizard/account_validate_account_move.py#L22
therefore method `_post` should be extended, and not `action_post`, to the warning is displayed when several invoices are validated

carlosdauden: Ping @psugne https://github.com/OCA/credit-control/commit/5d95c65e3c569426696b7d0400d51566e4105e14#diff-75cb51ce626ac1a0a6744302a700a18a6df917b7f76a8bb9a27a4d363c28e8bbR62

AaronHForgeFlow: thank you @carlosdauden that was a mistake doing the cherrypick. I included 4bd8fcd now. But I am setting this to WIP, I have to test it again.

account_financial_risk/models/res_partner.py
```diff
@@ -30,80 +31,80 @@ class ResPartner(models.Model):
...
-        string="Include Other Account Unpaid Amount",
+        string="Include Other Overdue Amounts",
         help="Full risk computation.\n"
         "Residual amount of move lines not reconciled with distinct "
         "account that is set as partner receivable and date maturity "
         "exceeded, considering Due Margin set in account settings.",
     )
     risk_account_amount_unpaid_limit = fields.Monetary(
-        string="Limit Other Account Unpaid Amount",
+        string="'Limit Other Overdue Amounts",
```
MiquelRForgeFlow, line 101:
```suggestion
        string="Limit Other Overdue Amounts",
```

AaronHForgeFlow:
This is actually a customization. Closing.

PR #161, opened 2021-11-30
[14.0][IMP/FIX] account_finacial_risk: add improvements

MiquelRForgeFlow:
First commit is a forward port.

PR #162, opened 2021-12-01
[15.0][IMP] account_financial_risk: forward port fixes

MiquelRForgeFlow:
Forward port of https://github.com/OCA/credit-control/pull/161.

### DIFF
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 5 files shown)


MiquelRForgeFlow:
@carlosdauden :)

carlosdauden (approved):

account_financial_risk/models/account_invoice.py
```diff
@@ -59,27 +60,39 @@ def risk_exception_msg(self):
...
         return exception_msg
 
-    def action_post(self):
+    def _post(self, soft=True):
         if (
             self.env.context.get("bypass_risk", False)
             or self.company_id.allow_overrisk_invoice_validation
         ):
-            return super().action_post()
+            return super()._post(soft)
```
pedrobaeza: Please always past keyword arguments as such:

```suggestion
            return super()._post(soft=soft)
```

MiquelRForgeFlow: but `_post` only has one parameter, it seems doing this is irrelevant :S

AaronHForgeFlow (approved):
Code review LGTM

PR #114, opened 2020-12-11
[FIX] sale_financial_risk: Expected singleton: sale.order()

BT-aestebanez:
Fixes #98

PR #163, opened 2021-12-08
[FIX] sale_financial_risk: Expected singleton: sale.order()

rousseldenis:
Forwardport of : https://github.com/OCA/credit-control/pull/114

### DIFF
--- a/sale_financial_risk/models/sale.py
+++ b/sale_financial_risk/models/sale.py
@@ -9,6 +9,7 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     def evaluate_risk_message(self, partner):
+        self.ensure_one()
         risk_amount = self.currency_id._convert(
             self.amount_total,
             self.company_id.currency_id,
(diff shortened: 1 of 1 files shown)


sale_financial_risk/models/sale.py
```diff
@@ -33,21 +34,22 @@ def evaluate_risk_message(self, partner):
...
-                    self.env["partner.risk.exceeded.wiz"]
-                    .create(
-                        {
-                            "exception_msg": exception_msg,
-                            "partner_id": partner.id,
-                            "origin_reference": "{},{}".format("sale.order", self.id),
-                            "continue_method": "action_confirm",
-                        }
+            for order in self:
+                partner = self.partner_invoice_id.commercial_partner_id
```
naglis: Wondering if we should take it from `order` instead of `self`?
```suggestion
                partner = order.partner_invoice_id.commercial_partner_id
```

naglis (approved):

carlosdauden (approved):

sale_financial_risk/models/sale.py
pedrobaeza: Please change this, @rousseldenis

pedrobaeza (approved):

pedrobaeza:
Cherry-pick to v15 at ec0b05b6acd773bbdedb95091b436faaff5380a9

PR #166, opened 2022-01-31
[MIG] account_credit_control: migration from version 10.0 to 14.0

fredzamoabg:
PR to migrate `account_credit_control` module `data` from version `10.0` to `14.0`


pedrobaeza:
This is not correct. You should cover migration scripts version per version.

PR #170, opened 2022-02-15
[IMP] account_invoice_overdue_reminder: allow to remove some undesired invoices on the fly during reminder process

legalsylvain:
**Summary**
This PR improve the module ``account_invoice_overdue_reminder`` to allow user to remove manually one or many invoices. This will recompute automatically the field ``counter``, and let user manually regenerate body and subject mail. (if reminder type is email)

**Use Case**
- Your customer has several invoices to pay. For one invoice, there is a pending conflict (customer think that the delivered products or the service is not ok, ...) You want to send a reminder without the conflictual invoice.
- Your customer has several invoices to pay. One of the invoice has been confirmed and sent late from your part. So you don't want to send a reminder for this invoice, but for the others.

**Screenshots**

**Step 1** : The wizard display all the invoices. The user can remove on or many invoices.

image

**Step2 ** : when an invoice is removed, counter is refreshed and an message is displayed to recompute email info. (title & body)

image

**Step 3** : Once clicked, the email info are updated. 

image

**Technical point**
- ``counter`` can be recomputed on the fly with a classic ``onchange`` function. However, recompute body is not possible in an 
- ``onchange`` function, because ``self.id`` is ``Newid`` in this context, making not possible the call of the function ``_render_template(..., res_id)``. For that reason a button is displayed.

Perhaps such workaround may be removed in more recent version of Odoo (with the refactor of onchange / depends core features).

### DIFF
--- a/account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
+++ b/account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
@@ -298,7 +298,7 @@ class OverdueReminderStep(models.TransientModel):
         'res.users', string='Assigned to', default=lambda self: self.env.user)
     letter_printed = fields.Boolean(readonly=True)
     invoice_ids = fields.Many2many(
-        'account.invoice', string='Overdue Invoices', readonly=True)
+        'account.invoice', string='Overdue Invoices')
     company_id = fields.Many2one(
         'res.company', readonly=True, required=True,
         default=lambda self: self.env['res.company']._company_default_get())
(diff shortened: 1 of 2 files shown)


legalsylvain:
Hi @alexis-via : Following my phone call this morning, here is a proposal from PR.

Thanks for your review !

legalsylvain:
@alexis-via @OCA/accounting-maintainers : could you take a look on that one. It is working well and in production since 4 monthes.

sergiocorato (approved):
LGTM

PR #174, opened 2022-03-06
[14.0] FIxes sale_financial_risk: Don't duplicate settings box

gdgellatly:
Before this commit you got a slighly offset layout in the config view.

After this layout, it is nice and standard.

### DIFF
--- a/sale_financial_risk/views/res_config_settings.xml
+++ b/sale_financial_risk/views/res_config_settings.xml
@@ -12,19 +12,17 @@
                 expr="//field[@name='allow_overrisk_invoice_validation']/../.."
                 position="inside"
             >
-                <div class="row col-xs-12 o_setting_box">
-                    <div class="o_setting_left_pane">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Maybe a screenshot of before and after?

gdgellatly:
Well here they are - but it should be obvious that you can't put a setting box in a setting box.
image

image

pedrobaeza (approved):

pedrobaeza:
And you get a fast-tracking as bonus for showcasing the problem :wink: 

/ocabot merge minor

pedrobaeza:
Cherry-picked whole feature with the patch included to v15 at 22ea4f6b83ffc0005fe94a57f0501117c1928d64. Previously, I needed to fix tests, as they were not being reported as incorrect at dd205ccb01250d49d2c58e473bd2c4c5d013f20f

PR #176, opened 2022-03-24
[14.0][FIX] account_financial_risk: inherit action_post instead of _post

fredzamoabg:
`_post` method return records and other modules could use it to perform any actions on this.

Based implementation from version `13.0` [here](https://github.com/OCA/credit-control/blob/2846f7f74ffab2e739f709e9e940d2dfa873c0d7/account_financial_risk/models/account_invoice.py#L62) that uses **post** method and the new mechanism in `14.0` in `account` module, is expected to inherit `action_post` insted of `_post` also seeing [this](https://github.com/odoo/odoo/blob/5b099bf6135b2ff517078d1839b513da7830f718/addons/account/models/account_move.py#L2575-L2581)

note: without this changes `account_financial_risk` will never show wizard because `_post` is encapsulated in `action_post` that is not interested in value returned from it
see https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L2716-L2718

### DIFF
--- a/account_financial_risk/tests/test_account_financial_risk.py
+++ b/account_financial_risk/tests/test_account_financial_risk.py
@@ -126,7 +126,7 @@ class TestPartnerFinancialRisk(SavepointCase):
         invoice2 = self.invoice.copy({"partner_id": self.invoice_address.id})
         self.assertAlmostEqual(self.partner.risk_invoice_draft, 550.0)
         self.assertAlmostEqual(self.partner.risk_invoice_unpaid, 550.0)
-        wiz_dic = invoice2._post()
+        wiz_dic = invoice2.action_post()
         wiz = self.env[wiz_dic["res_model"]].browse(wiz_dic["res_id"])
         self.assertEqual(wiz.exception_msg, "Financial risk exceeded.\n")
         self.partner.risk_invoice_unpaid_limit = 0.0
@@ -135,7 +135,7 @@ class TestPartnerFinancialRisk(SavepointCase):
         self.assertIn(self.partner, unrisk_partners)
         self.partner.risk_invoice_open_limit = 300.0
         invoice2.invoice_date_due = fields.Date.today()
-        wiz_dic = invoice2._post()
+        wiz_dic = invoice2.action_post()
         wiz = self.env[wiz_dic["res_model"]].browse(wiz_dic["res_id"])
         self.assertEqual(
             wiz.exception_msg, "This invoice exceeds the open invoices risk.\n"
@@ -144,7 +144,7 @@ class TestPartnerFinancialRisk(SavepointCase):
         self.partner.risk_invoice_draft_include = False
         self.partner.risk_invoice_open_include = True
         self.partner.credit_limit = 900.0
-        wiz_dic = invoice2._post()
+        wiz_dic = invoice2.action_post()
         wiz = self.env[wiz_dic["res_model"]].browse(wiz_dic["res_id"])
         self.assertEqual(
             wiz.exception_msg, "This invoice exceeds the financial risk.\n"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This was calling `_post`, not `post`, so I don't understand the problem.

fredzamoabg:
> This was calling `_post`, not `post`, so I don't understand the problem.

the problem is that _post returns records but this module use it to return dictionary (https://github.com/OCA/credit-control/blob/f12025fad55bd31adb66446b1433939777ed78ba/account_financial_risk/models/account_invoice.py#L82-L95)

so other modules could fail to use return value.

the correct method to be inherited is **action_post**

pedrobaeza:
OK, I understand, but tests are failing, so this makes something incorrect (or maybe it's showcasing a problem in the design of the tests, calling twice to post).

fredzamoabg:
> OK, I understand, but tests are failing, so this makes something incorrect (or maybe it's showcasing a problem in the design of the tests, calling twice to post).

yes I'll check, thank you

account_financial_risk/tests/test_account_financial_risk.py
```diff
@@ -210,20 +210,16 @@ def test_recompute_newid(self):
...
         invoice2 = self.invoice.copy({"partner_id": self.invoice_address.id})
-        validate_wiz = (
-            self.env["validate.account.move"]
-            .with_context(active_model="account.move", active_ids=invoice2.ids)
-            .create({})
-        )
         with self.assertRaises(ValidationError):
-            invoice2._post()
-            validate_wiz.validate_move()
+            invoice2.with_context(
```
pedrobaeza: This is not doing the same, as we want to try the batch move validation.

fredzamoabg: Uhm ok... there is another way you know for doing the same? As this wizard works with `_post` method

pedrobaeza: @carlosdauden you are the one that know this more. Can you give your advice?

fredzamoabg:
@pedrobaeza I updated PR description with main problem for this module

account_financial_risk/tests/test_account_financial_risk.py
carlosdauden: Yes, keep in mind that the restriction must work both from the form view button and when multiple invoices are selected from the list view. Odoo doesn't offer the best implementations to extend methods...
I'll take a look at it next week.

fredzamoabg: the only way is to inherit both `action_post` and `_post`... I'll try and back to you asap

fredzamoabg: @pedrobaeza @carlosdauden I changed implementation to cover invoice confirmation from both form and list.

tafaRU (approved):

carlosdauden (approved):
Code LGTM

pedrobaeza:
Thanks for hunting this until the resolution:

/ocabot merge patch

Is this happening in v13?

fredzamoabg:
> Thanks for hunting this until the resolution:
> 
> /ocabot merge patch
> 
> Is this happening in v13?

I don't think so, as `post` method in `v13` is returning `boolean` https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L2382

pedrobaeza:
Thanks for checking. I'm seeing that v15 is also migrated, so I suppose this one is needed, isn't it?

fredzamoabg:
> Thanks for checking. I'm seeing that v15 is also migrated, so I suppose this one is needed, isn't it?

yes (same behavior like 14.0)

pedrobaeza:
Can you please do the PR?

fredzamoabg:
> Can you please do the PR?

yes of course, I'll do it asap

PR #181, opened 2022-04-07
[13.0] account_invoice_overdue_reminder: add migration scripts for v12 to v13

alexis-via:


account_invoice_overdue_reminder/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.table_exists(env.cr, "account_invoice"):
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE account_invoice_overdue_reminder aior
+            SET invoice_id = account_invoice.move_id
+            FROM account_invoice
+            WHERE aior.%(old_inv_id)s = account_invoice.id"""
```
MiquelRForgeFlow, line 17:
```suggestion
            UPDATE account_invoice_overdue_reminder aior
            SET invoice_id = am.id
            FROM account_move am
            WHERE aior.%(old_inv_id)s = am.old_invoice_id"""
```
if you are using OpenUpgrade project, this is better

MiquelRForgeFlow (commented):
pre-commit needs to be fixed

PR #184, opened 2022-04-12
[12.0] fix account_credit_control migration script

victor-champonnois:
The 12.0.3.0.4 pre-migration script raises a `ValueError` when migrating from a version below 13.0.3.0. This is because the `credit_control_communication_action` action has not been created yet (it is created in 12.0.3.0).

### DIFF
--- a/account_credit_control/migrations/12.0.3.0.4/pre-migration.py
+++ b/account_credit_control/migrations/12.0.3.0.4/pre-migration.py
@@ -11,5 +11,9 @@ def migrate(env, version):
     This patch release fixes a bug whose result may be stored in database, so
     although it's still a patch, we need to clear the database, just in case.
     """
-    action = env.ref("account_credit_control.credit_control_communication_action")
-    action.domain = []
+    action = env.ref("account_credit_control.credit_control_communication_action",
+                     raise_if_not_found=False)
+    # if not found, this XML id has not been created,
+    # so clearing the database is not needed.
+    if action:
+        action.domain = []


remytms:
To explain a bit more, the fix in the pre-migration script has no meaning if the record `credit_control_communication_action` is not already in the database. So it should only run if this action is already present in the database.

account_credit_control/migrations/12.0.3.0.4/pre-migration.py
```diff
@@ -11,5 +11,9 @@ def migrate(env, version):
     This patch release fixes a bug whose result may be stored in database, so
     although it's still a patch, we need to clear the database, just in case.
     """
-    action = env.ref("account_credit_control.credit_control_communication_action")
-    action.domain = []
+    try:
+        action = env.ref("account_credit_control.credit_control_communication_action")
```
sergiocorato: Better use `, raise_if_not_found=False)` instead of try-except

victor-champonnois: fixed, thanks :)

account_credit_control/migrations/12.0.3.0.4/pre-migration.py
```diff
@@ -11,5 +11,6 @@ def migrate(env, version):
     This patch release fixes a bug whose result may be stored in database, so
     although it's still a patch, we need to clear the database, just in case.
     """
-    action = env.ref("account_credit_control.credit_control_communication_action")
+    action = env.ref("account_credit_control.credit_control_communication_action", raise_if_not_found=False)
+    # if not found, this XML id has not been created, so clearing the database is not needed.
     action.domain = []
```
sergiocorato: `action` can be `False`, so it need to be checked before assign values

victor-champonnois: it's fixed

sergiocorato (approved):
LGTM

sergiocorato (changes requested):
Please fix Travis error

victor-champonnois:
Travis fixed, the codecov/project does not pass but I am not sure if I have to do something about it.

remytms:
@sergiocorato Travis has been fixed and is green. :) Can you review again ? Thanks

sergiocorato (approved):
Thanks!

sergiocorato:
(It needs another approval to merge)

carlosdauden (approved):

PR #188, opened 2022-05-09
[FIX] account_credit_control: Allow to use run on multicompany environments

etobella:
With the change, run works properly if no company is set (generated for all visible companies)

### DIFF
--- a/account_credit_control/models/credit_control_run.py
+++ b/account_credit_control/models/credit_control_run.py
@@ -132,14 +132,15 @@ class CreditControlRun(models.Model):
 
             if policy.do_nothing:
                 continue
-            (
-                policy_manual_lines,
-                policy_lines_generated,
-                policy_report,
-            ) = policy._generate_credit_lines(self, {"run_id": self.id})
-            manually_managed_lines |= policy_manual_lines
-            generated |= policy_lines_generated
-            report += policy_report
+            for company in self.company_id or self.env.companies:
+                (
+                    policy_manual_lines,
+                    policy_lines_generated,
+                    policy_report,
+                ) = policy._generate_credit_lines(self, company, {"run_id": self.id})
+                manually_managed_lines |= policy_manual_lines
+                generated |= policy_lines_generated
+                report += policy_report
 
         vals = {
             "state": "done",
--- a/account_credit_control/models/credit_control_communication.py
+++ b/account_credit_control/models/credit_control_communication.py
@@ -114,15 +114,15 @@ class CreditControlCommunication(models.Model):
     ):
         """ Return credit lines related to a partner and a policy level """
         cr_line_obj = self.env["credit.control.line"]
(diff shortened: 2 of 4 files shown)


etobella:
@pedrobaeza May I ask for your review?

account_credit_control/report/report_credit_control_summary.xml
```diff
@@ -53,7 +53,7 @@
                             <td>
                                 <span t-field="l.date_due" />
                             </td>
-                            <td class="text-right">
+                            <td class="text-righcbt">
```
pedrobaeza: Is this change intended?

account_credit_control/models/credit_control_run.py
```diff
@@ -132,14 +132,15 @@ def _generate_credit_lines(self):
...
                 continue
-            (
-                policy_manual_lines,
-                policy_lines_generated,
-                policy_report,
-            ) = policy._generate_credit_lines(self, {"run_id": self.id})
-            manually_managed_lines |= policy_manual_lines
-            generated |= policy_lines_generated
-            report += policy_report
+            for company in self.company_id or self.env.companies:
```
pedrobaeza, line 135: We shouldn't include in the same run several companies, as their records will be only visible if all companies are allowed at the same time.

pedrobaeza (changes requested):

account_credit_control/report/report_credit_control_summary.xml
etobella: Probably a typo error from my side. Sorry :disappointed:

account_credit_control/models/credit_control_run.py
etobella: Well, right now the field `company_id` on the credit.control.run is not required. Users expects that if the field is null, it will create the credit lines for all the companies visible :thinking: 
We can make it required and forget about making this multicompany. Also we could make the field M2M in order to avoid doubts. What do you prefer?

pedrobaeza: I think it should be done mandatory. There's no sense of executing a multi-company run that mixes up everything and to have to make the code aware of that. Imagine having the same partner with debts in several companies.

Unless it's your use case that you want to cover in one shot...

etobella: The idea was to generate all the lines on the same execution

But, data is not mixed, communications and credit lines are created by company and partner, not only by partner, so they are not mixed.

pedrobaeza: They all belong to the same run, which is the inconsistent thing.

etobella: It is not if it has no company or multiple company associated

pedrobaeza: Yes. The problem is the or party. Make required the field company_id

PR #193, opened 2022-06-14
[13.0][IMP] account_invoice_overdue_reminder: add method to get reminder template

JudithAforgeFlow:
In case you want to use more than one different templates for the reminders it is more convenient to have a dedicated method.

CC @ForgeFlow


JordiMForgeFlow (approved):
LGTM

MiquelRForgeFlow (approved):

carlosdauden (approved):

PR #194, opened 2022-06-14
[14.0][IMP] sale_financial_risk_info: apply currency format

enriquemartin:
Apply the currency format for both the symbol and the number of decimal places.


carlosdauden (approved):

pedrobaeza:
And what happen with already existing ICP that contains `{symbol}` in its definition?

pedrobaeza (approved):

PR #204, opened 2022-07-28
[13.0][IMP] sale_payment_sheet_financial_risk: Reduce each risk field depending…

carlosdauden:
… of payment sheet invoice status (type of debt)

@Tecnativa TT37610


sale_payment_sheet_financial_risk/models/res_partner.py
```diff
@@ -21,15 +23,20 @@ class ResPartner(models.Model):
...
     )
     sale_payment_sheet_ids = fields.One2many(
         comodel_name="sale.payment.sheet.line", inverse_name="partner_id"
     )
 
     def _get_risk_sale_payment_sheet_domain(self):
-        # When p is NewId object instance bool(p.id) is False
         commercial_partners = self.filtered(
-            lambda p: (p.id and p == p.commercial_partner_id)
+            lambda p: (p.ids == p.commercial_partner_id.ids)
```
pedrobaeza, line 40: The order is assured to be the same always? If not, do:

```suggestion
            lambda p: (set(p.ids) == set(p.commercial_partner_id.ids))
```

carlosdauden: p in lambda filtered is always only 1 record

carlosdauden: And commercial_partner_id is a Many2One field that only contains 1 record

pedrobaeza: Then why not using `p == p.commercial_partner_id`?

carlosdauden: When p is NewId object instance is not equal to "Normal" registry

Modified explain message:
https://github.com/OCA/credit-control/pull/204/files#diff-9407b845adfe69cef43bd1306655f899ac7ac40987e58638425c2790365ae459R38

pedrobaeza: OK then

pedrobaeza (approved):

PR #202, opened 2022-07-18
[12.0] IMP account_invoice_overdue_reminder adding "other" reminder type

eLBati:

PR #206, opened 2022-08-23
[IMP] account_invoice_overdue_reminder: Add "Other" Reminder Type

SirTakobi:
Supersede https://github.com/OCA/credit-control/pull/202.

### DIFF
--- a/account_invoice_overdue_reminder/models/overdue_reminder_action.py
+++ b/account_invoice_overdue_reminder/models/overdue_reminder_action.py
@@ -46,6 +46,7 @@ class OverdueReminderAction(models.Model):
             ('mail', _('E-mail')),
             ('phone', _('Phone')),
             ('post', _('Letter')),
+            ('other', _('Other')),
             ]
 
     @api.depends('reminder_ids')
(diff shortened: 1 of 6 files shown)


account_invoice_overdue_reminder/models/overdue_reminder_action.py
```diff
@@ -46,6 +50,7 @@ def _reminder_type_selection(self):
             ('mail', _('E-mail')),
             ('phone', _('Phone')),
             ('post', _('Letter')),
+            ('other', _('Other')),
```
legalsylvain: Other seems to generic for me. Something like 'manual' looks better. Don't you think ? 

same for the function, so validate_other --> validate_manual.

legalsylvain (commented):
One remarks. Otherwise, LGTM. 
Thanks !

account_invoice_overdue_reminder/models/overdue_reminder_action.py
SirTakobi: Thanks for the review!
I have changed the code as you suggested, can you have another look?

legalsylvain (approved):
Thanks !

alexis-via:
Just for my understanding: what kind of real-life scenario do you put under this new type manual/other ?

SirTakobi:
> Just for my understanding: what kind of real-life scenario do you put under this new type manual/other ?

Asked this to the functional guys, they say that this can be used when the salesman goes see the customer, or the salesman leaves a message to the customer.
cc @eLBati

mrcast (approved):
Functional test, ok!

SirTakobi:
@OCA/accounting-maintainers this has been approved after the bot marked it as stale, can someone reopen it? I can't :(
Thanks!

PR #207, opened 2022-08-24
[14.0][IMP] sale_financial_risk_info: Absolute and percentage values of re…

ilyasProgrammer:
…maining risk added to partner and Sale Order as computed fields to make it compatible with domain widget.

### DIFF
--- a/sale_financial_risk_info/models/sale.py
+++ b/sale_financial_risk_info/models/sale.py
@@ -9,6 +9,10 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     risk_info = fields.Html(compute="_compute_risk_info")
+    risk_remaining_value = fields.Monetary(related="partner_id.risk_remaining_value")
+    risk_remaining_percentage = fields.Float(
+        related="partner_id.risk_remaining_percentage"
+    )
 
     @api.depends("partner_invoice_id")
     def _compute_risk_info(self):
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -146,6 +146,31 @@ class ResPartner(models.Model):
     risk_currency_id = fields.Many2one(
         comodel_name="res.currency", compute="_compute_credit_currency"
     )
+    risk_remaining_value = fields.Monetary(
+        compute="_compute_risk_remaining",
+        string="Risk Remaining (Value)",
+        currency_field="risk_currency_id",
+        store=True,
+    )
+
+    risk_remaining_percentage = fields.Float(
+        compute="_compute_risk_remaining",
+        string="Risk Remaining (Percentage)",
+        store=True,
+    )
+
+    @api.depends("credit_limit", "risk_total")
+    def _compute_risk_remaining(self):
+        for record in self:
+            record.risk_remaining_value = record.credit_limit - record.risk_total
+            if record.credit_limit:
+                record.risk_remaining_percentage = round(
+                    100
+                    * (record.credit_limit - record.risk_total)
+                    / record.credit_limit
+                )
+            else:
+                record.risk_remaining_percentage = 0
 
     @api.depends(
         "credit_currency",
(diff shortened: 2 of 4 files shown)


francesco-ooops (approved):
Functional review ok 

image

now it's possible to use these fields for approval tiers, automated actions, etc.

elvise (approved):
Functional review ok!

pedrobaeza:
Part of this must be in `account_financial_risk`, not in `sale_financial_risk`.

carlosdauden (changes requested):
Extend https://github.com/OCA/credit-control/tree/14.0/sale_financial_risk_info
or create a new module to avoid penalizing performance in installations that do not need it

francesco-ooops (approved):
functional ok

@carlosdauden @pedrobaeza better now?

francesco-ooops:
@carlosdauden could you take a look? thanks :)

sale_financial_risk_info/models/sale.py
```diff
@@ -9,6 +9,10 @@ class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     risk_info = fields.Html(compute="_compute_risk_info")
+    risk_remaining_value = fields.Monetary(related="partner_id.risk_remaining_value")
+    risk_remaining_percentage = fields.Float(
+        related="partner_id.risk_remaining_percentage"
+    )
```
carlosdauden, line 15: You can use these new fields in _compute_risk_info:
https://github.com/OCA/credit-control/blob/14.0/sale_financial_risk_info/models/sale.py#L26

https://github.com/OCA/credit-control/blob/14.0/sale_financial_risk_info/models/sale.py#L39-L44

carlosdauden (approved):

sale_financial_risk_info/models/sale.py
ilyasProgrammer: risk_available and risk_remaining_percentage have different formulas.

carlosdauden: Ok, in that case you can include the risk_remaining_percentage parameter in `format info_pattern.format(`

risk_available and risk_remaining_value have the same operation `record.credit_limit - record.risk_total`

This is just a suggestion, I already approved the PR

francesco-ooops:
@pedrobaeza merge?

sale_financial_risk_info/models/res_partner.py
```diff
@@ -1,10 +1,36 @@
...
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
     risk_percent_warning = fields.Integer(string="Risk percent warning", default=80)
+
+    risk_remaining_value = fields.Monetary(
```
pedrobaeza: I still think these fields shouldn't go in any `sale_financial_*` module, as they are general values

pedrobaeza (changes requested):

sale_financial_risk_info/models/res_partner.py
francesco-ooops: @pedrobaeza in which module do you suggest we could create these fields?

francesco-ooops: @pedrobaeza when you can, if you coul provide directions we'll try to close this PR :)

pedrobaeza:
I think this should go to `account_financial_risk`, as it's general information, not only sales related one. In fact, visually I think they should go on the left part above total risk information. It can be also compacted in one line to show both data.

Not sure if Carlos' fear is that this affects performance, but I don't think so and even he points to a method that do everything in one shot.

ilyasProgrammer:
updated

carlosdauden:
Let's merge and if there is a performance penalty we analyze it

francesco-ooops:
@pedrobaeza go? :)

francesco-ooops:
@pedrobaeza good for merge?

account_financial_risk/static/src/scss/from_view.scss
```diff
@@ -0,0 +1,24 @@
+.o_form_view {
```
pedrobaeza: Can you please avoid this extra CSS? I think using `<div>` and `class="oe_inline"` you can do the same.

francesco-ooops: updated

pedrobaeza (approved):

PR #208, opened 2022-08-24
[15.0][ADD] dunning_block: temporarily suspends the dunning process

JakubWiselka:
If you want to postpone the dunning process for some invoices you
can activate dunning block on them in credit control page
or through the action. You need to choose the end date of dunning block,
reason and short note. Invoices with active dunning block will not
be taken into the credit control run. And after given date dunning block
will be automatically disabled and invoice will return to credit control run.

It might be useful when f.e customer says that he cannot pay until end of the month anyway, and you don't want to remind him about payment every week, but still want to keep track this case automatically


dunning_block/models/account_move.py
```diff
@@ -0,0 +1,99 @@
...
+        readonly=True, store=True, tracking=True, copy=False
+    )
+    dunning_block_history_ids = fields.One2many(
+        "dunning.block.history", "move_id", readonly=True, copy=False
+    )
+    dunning_block_days_left = fields.Integer(compute="_compute_dunning_block_days_left")
+    dunning_block_reason_id = fields.Many2one(
+        "dunning.block.reason", readonly=True, store=True, tracking=True, copy=False
+    )
+
```
carlosdauden, line 37: Instead of overloading the stock.move model it would be better to use dunning.block.history (or another similar model) to store the blocking information and query directly from that list which invoices should be dunned.

carlosdauden (changes requested):

dunning_block/models/account_move.py
JakubWiselka: Totally agree, would be better to have it another model and maybe just link it with singe many2one field. I could maybe rewrite it if there would be an interest in this functionality

PR #209, opened 2022-08-29
[15.0][FIX]sale_financial_risk: Invoice from Sales Order should show risk confirmation wizard

AaronHForgeFlow:
If you go to a sales order and then to the invoices from the button and you try to validate the invoice you currently get the exception message, but you are not able to continue with the validation even if you are a partner risk manager.

However, if you go to the same invoice from the accounting menu you will be able to validate the invoice, because the context is not passed. This sentence here in the account_financial_risk module is not correct when you come from sales orders: https://github.com/OCA/credit-control/blob/15.0/account_financial_risk/models/account_invoice.py#L102

In my opinion it should work the same way, as you are trying to validate the same invoice. WDYT?

cc @ForgeFlow @MiquelRForgeFlow

### DIFF
--- a/sale_financial_risk/models/__init__.py
+++ b/sale_financial_risk/models/__init__.py
@@ -2,3 +2,4 @@ from . import payment
 from . import res_partner
 from . import sale
 from . import res_config_settings
+from . import account_invoice
--- a/sale_financial_risk/models/account_invoice.py
+++ b/sale_financial_risk/models/account_invoice.py
@@ -1,0 +1,31 @@
+# Copyright 2016-2018 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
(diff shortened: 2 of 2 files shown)


JordiMForgeFlow (approved):
Code LGTM and makes sense 👍🏼

carlosdauden:
I agree that it should work the same way, but I'm not sure if this is the best solution.
I am investigating an alternative directly in account_financial_risk.

carlosdauden:
Try https://github.com/OCA/credit-control/pull/223 to solve this issue

MiquelRForgeFlow:
Should this be closed, then?

carlosdauden:
Yes, if the alternative solution seems appropriate to you :wink:

pedrobaeza:
The alternative is already merged, so closing.

pedrobaeza:
Please comment if there's still any non covered case.

PR #221, opened 2022-10-04
[15.0][FIX] account_financial_risk: issue with currenty rates.

LoisRForgeFlow:
When doing conversions of currencies we cannot ignore the invoice dates, otherwise differences between de residual currency amount of the invoices and the computed risk amounts can arise.

Steps to reproduce:

1. Create 2 rates for a secondary currency.
2. Enter an invoice in that secondary currency and set the invoice date in the past so it uses the oldest rate.
3. Check financial risk fields.

The fix implemented is a compromise. On one hand, having the most robust fix would imply iterate all invoices individually. On the other hand, loosing the performance benefits given by the `read_group` is also not desired.

@ForgeFlow

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -230,6 +230,13 @@ class ResPartner(models.Model):
     def _risk_account_groups(self):
         max_date = self._max_risk_date_due()
         company_domain = self._get_risk_company_domain()
+        fields = [
+            "partner_id",
+            "account_id",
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
This solution is a good for most cases and do not compromise performance so :+1:

JordiMForgeFlow (approved):
Code review LGTM 👍🏼

MiquelRForgeFlow (approved):

PR #227, opened 2022-10-31
[15.0][IMP] account_financial_risk: credit limit last modification date field.

angelgarciadelachica:
A new field is created that stores the date of the last credit limit modification if the credit limit is non-zero.
In case the credit limit is non-zero, the date of the last modification is displayed next to the credit limit in the customer's Finacial Risk tab.
Captura 01

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -1,6 +1,8 @@
 # Copyright 2016-2018 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from datetime import datetime
+
 from dateutil.relativedelta import relativedelta
 
 from odoo import _, api, fields, models
@@ -145,6 +147,19 @@ class ResPartner(models.Model):
     risk_currency_id = fields.Many2one(
         comodel_name="res.currency", compute="_compute_credit_currency"
     )
+    date_credit_limit = fields.Date(
+        compute="_compute_date_credit_limit",
+        store=True,
+        readonly=False,
+        string="Last Credit Limit Date",
+    )
+
+    @api.depends("credit_limit")
+    def _compute_date_credit_limit(self):
+        self.filtered(lambda x: x.credit_limit == 0.00).date_credit_limit = False
+        self.filtered(
+            lambda x: x.credit_limit != 0.00
+        ).date_credit_limit = datetime.today()
 
     @api.depends(
         "credit_currency",
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,9 +1,14 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.5.2
+_commit: v1.11.0
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
 dependency_installation_mode: PIP
 generate_requirements_txt: true
+github_check_license: true
+github_enable_codecov: true
+github_enable_makepot: true
+github_enable_stale_action: true
+github_enforce_dev_status_compatibility: true
 include_wkhtmltopdf: false
 odoo_version: 15.0
 org_name: Odoo Community Association (OCA)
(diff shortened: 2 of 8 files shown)


angelgarciadelachica:
@ValentinVinagre @HaraldPanten

pedrobaeza:
The field is already tracked:

https://github.com/OCA/credit-control/blob/9e5ecdfa93ab2d8207792d70b979d6aef5c5f5dc/account_financial_risk/models/res_partner.py#L132

and thus, each change is written on the chatter with the date, so you have the full change history.

I think that adding overhead for the same feature doesn't worth.

ValentinVinagre:
> The field is already tracked:
> 
> https://github.com/OCA/credit-control/blob/9e5ecdfa93ab2d8207792d70b979d6aef5c5f5dc/account_financial_risk/models/res_partner.py#L132
> 
> and thus, each change is written on the chatter with the date, so you have the full change history.
> 
> I think that adding overhead for the same feature doesn't worth.

It is true that there is a tacking, but when you try to filter those that indicated the credit limit for more than 6 months or 1 year to request it again, you cannot see that information. In this way if it would be possible and it is hidden next to the maximum.

pedrobaeza:
I don't see that a great problem, but if others think that the overhead is worth, then go ahead.

ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@HaraldPanten could you check it?

account_financial_risk/models/res_partner.py
```diff
@@ -145,6 +147,21 @@ class ResPartner(models.Model):
     risk_currency_id = fields.Many2one(
         comodel_name="res.currency", compute="_compute_credit_currency"
     )
+    date_credit_limit = fields.Date(
+        compute="_compute_date_credit_limit",
+        store=True,
+        readonly=True,
```
carlosdauden: Maybe this field should be writable. (keeping the rest of the functionality)
What happens if we want to save the date in which we have changed the conditions and that date has already passed?
In Spain, it is quite common for the administration department to introduce the modifications with some delay. :wink: 
```suggestion
        readonly=False,
```

account_financial_risk/models/res_partner.py
```diff
@@ -145,6 +147,21 @@ class ResPartner(models.Model):
...
+        string="Last Credit Limit Date",
+    )
+
+    @api.depends("credit_limit")
+    def _compute_date_credit_limit(self):
+        self.filtered(lambda x: x.credit_limit == 0.00).write(
+            {"date_credit_limit": None}
+        )
```
carlosdauden:
```suggestion
        self.filtered(lambda x: x.credit_limit == 0.00).date_credit_limit = False
```

account_financial_risk/models/res_partner.py
```diff
@@ -145,6 +147,21 @@ class ResPartner(models.Model):
...
+        self.filtered(lambda x: x.credit_limit != 0.00).write(
+            {"date_credit_limit": datetime.today()}
+        )
```
carlosdauden:
```suggestion
        self.filtered(lambda x: x.credit_limit != 0.00).date_credit_limit = datetime.today()
```

carlosdauden (changes requested):

HaraldPanten:
@angelgarciadelachica once you check Carlos's comments, that will be OK for me.

carlosdauden (approved):
Code LGTM

pedrobaeza:
The repo needs a copier template update. Can you please do it here or in a separate PR?

HaraldPanten:
Hi @pedrobaeza that's the same as your team did here, right? --> https://github.com/OCA/l10n-spain/pull/2579

pedrobaeza:
Yes, it's executing `copier -f update` in the repo.

HaraldPanten:
THX! we'll have a look 👍

HaraldPanten (approved):

angelgarciadelachica:
It's done!
THX @pedrobaeza.
In case someone has the same problem I found the following link to fix it:
https://github.com/OCA/oca-addons-repo-template

carlosdauden:
Can you review this view in edition mode?

image

angelgarciadelachica:
Sure, I check it.

PR #232, opened 2022-11-17
[12.0][IMP] copier

sergiocorato:


legalsylvain (approved):
Thanks ! Maintainers could merge ?

pedrobaeza (approved):
All these efforts may go to migrate to upper versions, but not judging.

PR #237, opened 2022-11-29
[FIX] account_invoice_overdue_reminder : replace user id by user name (get with sudo)

legalsylvain:
To avoid having error access if the user of the last entry is not accessible by the current user

**Step to reproduce**

- Use this module in a multi company context.
- user A create an entry in the journal then switch the company
- user B want to use the module "overdue reminder". 
- An error is raised, because User B can not read the name of user B.


alexis-via:
I don't think the problem is in the code pointed out in this PR. If you can't read the name of any user in a multi-company context, then you have a lot of similar problem elsewhere. For me, in multi-company context, read access on res.users should be allowed.

For example, on an invoice or sale.order, you have a user_id field. If this user has switched to another company, you can't access the invoice or sale.order any more ?

legalsylvain:
make sense. I'll work on my multicompany design.
Thanks for your answer.

PR #243, opened 2023-01-16
[14.0][FIX] account_financial_risk - adapt exception msg to multi company

baimont:
Running the account's cron "Account; Post draft entries with auto_post set to True up to today" on a multi companies database, we got an Expected singleton error:

`ValueError: <class 'ValueError'>: "Expected singleton: res.company(1, 2)" while evaluating 'model._autopost_draft_entries()'`

that we traced back to this line.

### DIFF
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -64,12 +64,12 @@ class AccountMove(models.Model):
         Method used to return the first invoice with exception message.
         """
         ret = False, False
-        if (
-            self.env.context.get("bypass_risk", False)
-            or self.company_id.allow_overrisk_invoice_validation
-        ):
+        if self.env.context.get("bypass_risk", False):
             return ret
-        for invoice in self.filtered(lambda x: x.move_type == "out_invoice"):
+        for invoice in self.filtered(
+            lambda x: x.move_type == "out_invoice"
+            and not x.company_id.allow_overrisk_invoice_validation
+        ):
             exception_msg = invoice.risk_exception_msg()
             if exception_msg:
                 ret = invoice, exception_msg


account_financial_risk/models/account_invoice.py
```diff
@@ -64,9 +64,8 @@ def _first_invoice_exception_msg(self):
...
         """
         ret = False, False
-        if (
-            self.env.context.get("bypass_risk", False)
-            or self.company_id.allow_overrisk_invoice_validation
+        if self.env.context.get("bypass_risk", False) or all(
+            self.company_id.mapped("allow_overrisk_invoice_validation")
         ):
             return ret
         for invoice in self.filtered(lambda x: x.move_type == "out_invoice"):
```
carlosdauden: Try don't repeat loop
```suggestion
        if self.env.context.get("bypass_risk", False):
            return ret
        for invoice in self.filtered(lambda x: x.move_type == "out_invoice" and not x.company_id.allow_overrisk_invoice_validation):
```

carlosdauden (changes requested):

carlosdauden (approved):

FrancoMaxime (approved):

acsonefho (approved):
LGTM

acsonefho:
@OCA/accounting-maintainers if someone have super-power here to review/merge here!
Easy one!

AaronHForgeFlow (approved):

AaronHForgeFlow:
@baimont thanks for the fix. Can you check if it is needed in v15?

pedrobaeza:
Can you fw-port this to upper versions?

pedrobaeza:
Please someone do the forward-port

pedrobaeza:
Forward-ported on #303

PR #245, opened 2023-01-20
[16.0][MIG] sale_financial_risk: Migration to 16.0

BT-atornero:


BT-atornero:
has dependencies to ```account_financial_risk```

PR #246, opened 2023-01-23
[16.0][MIG] account_financial_risk: Migration to 16.0

BT-atornero:


BrunoBailo:
Hello, I have downloaded this module in branch 16 from the pull request. When I select accounting-customers-customers, then create or select one and dive into Financial Risk, it appears this way in my data base:
Captura desde 2023-02-09 13-04-56

PR #260, opened 2023-02-20
[15.0][FIX] account_credit_control

dsolanki-initos:
message_format() got an unexpected keyword argument 'format_reply'

### DIFF
--- a/account_credit_control/models/mail_message.py
+++ b/account_credit_control/models/mail_message.py
@@ -7,8 +7,8 @@ from odoo import models
 class MailMessage(models.Model):
     _inherit = "mail.message"
 
-    def message_format(self):
-        result = super().message_format()
+    def message_format(self, format_reply=True):
+        result = super().message_format(format_reply=format_reply)
         credit_control = self.env["ir.model.data"]._xmlid_to_res_id(
             "account_credit_control.mt_request"
         )


account_credit_control/models/mail_message.py
```diff
@@ -7,8 +7,8 @@
 class MailMessage(models.Model):
     _inherit = "mail.message"
 
-    def message_format(self):
-        result = super().message_format()
+    def message_format(self, format_reply=True):
+        result = super().message_format(format_reply)
```
pedrobaeza: Respect keyword arguments

```suggestion
        result = super().message_format(format_reply=format_reply)
```

pedrobaeza (changes requested):
Change also message commit to comply with guidelines:

`[FIX] account_credit_control: Params of message_format`

account_credit_control/models/mail_message.py
dsolanki-initos: @pedrobaeza , I've fixed the commit message and keyword arguments as you mentioned above. Please review it.

pedrobaeza (approved):
The fix is clear:

https://github.com/odoo/odoo/blob/c55badb8d70bdccfeb82bffb330d3f0b1b162ed1/addons/mail/models/mail_message.py#L911

so fast-tracking it:

/ocabot merge patch

PR #266, opened 2023-02-28
[16.0][MIG] account_invoice_overdue_warn

alexis-via:


legalsylvain:
Hi @alexis-via . thanks for porting this usefull module.
I use this module in V12 and for me the banner is missing in the invoice form view. Any reason not to put the banner also here ?
(I can do a PR if you're agree)
Thanks !

alexis-via:
@legalsylvain The banner in the invoice form view is added by the module account_invoice_overdue_reminder, cf https://github.com/OCA/credit-control/blob/14.0/account_invoice_overdue_reminder/views/account_move.xml#L16
So I would recommend to backport this warning banner in account_invoice_overdue_reminder.

legalsylvain:
Hi @alexis-via. Thanks for your answer. AFAIK : 
- the banner in the module ``account_invoice_overdue_reminder`` say that the current invoice is overdue.
- the banner present in the ``account_invoice_overdue_warn(_sale)`` display the total amount of the overdue with the number of invoices (that are very important informations). 

The context is that sometimes, people create directly sale invoices without creating sale order before. And at this step, having the same information as in the sale order is very valuable.

alexis-via:
OK, I understand your point, and I agree that it would be a nice improvement.
It means that, if the invoice itself is overdue, we would have 2 warning banners:
- the one from account_invoice_overdue_reminder which says that the invoice is overdue with the button "Send an overdue reminder"
- the one from account_invoice_overdue_warn which says that this partner has N invoice invoices for a total amount of xx,yy €
Is it a problem to have 2 warnings banners? I don't think it's a big problem, but it may be a bit "too much".
To avoid having 2 warning banners, my idea would be to use a ribbon flag "Overdue" (similar to the "Paid" ribbon) and put the button "Send an overdue reminder" at the top with the other buttons.
If you think it's a good idea, I'm ok to do it in v16.

legalsylvain:
>  Is it a problem to have 2 warnings banners? I don't think it's a big problem, but it may be a bit "too much".

yes, should be avoided if possible. it's not a big deal either.

> To avoid having 2 warning banners, my idea would be to use a ribbon flag "Overdue" (similar to the "Paid" ribbon) and put the button "Send an overdue reminder" at the top with the other buttons.

it will conflict with the current banner "paid" / "opened" / ... Don't you think ? I don't know odoo view that have many banners in the same time.

Another idea is to create a glue module. if installed, we hide the "more simple banner". Quite trivial glue module I think.

I don't have a very clear opinion on the best way to do.

Bearnard21 (approved):
Functional test: Everything performs well.

alexis-via:
@legalsylvain You're certainly right about the fact that it may not be possible to have 2 banners on the same document... I'll make some tests.
Another idea would be to integrate account_invoice_overdue_warn inside account_invoice_overdue_reminder. The only consequence it that you would not be able to use the "small" module "account_invoice_overdue_warn" alone without the "big" module "account_invoice_overdue_reminder". I don't know anybody who uses account_invoice_overdue_warn without account_invoice_overdue_reminder... but I don't pretend to know all the users of the modules I write!

legalsylvain:
> @legalsylvain You're certainly right about the fact that it may not be possible to have 2 banners on the same document... I'll make some tests.

OK, let me know when you have tested.

> Another idea would be to integrate account_invoice_overdue_warn inside account_invoice_overdue_reminder. The only consequence it that you would not be able to use the "small" module "account_invoice_overdue_warn" alone without the "big" module "account_invoice_overdue_reminder". I don't know anybody who uses account_invoice_overdue_warn without account_invoice_overdue_reminder

Even if both module are "often installed together", I think it's better to avoid forcing users to install a complex module, if they want a simple functionality.

Any way, it's not a blocking point for the current migration, and can be added in a second PR.

GabbasovDinar (approved):
Code LGTM

ivs-cetmix:
Hey @OCA/accounting-maintainers could you please have a look at this?

HaraldPanten:
@alexis-via will you continue with this current PR, and this other one? --> https://github.com/OCA/credit-control/pull/267

PR #269, opened 2023-02-28
[16.0][MIG] account_invoice_overdue_reminder

alexis-via:
Migrated the module with several small improvements/fixes:

- result_notes fields converted from fields.Text to fields.Html
- In mass view of steps, show buttons in tree view
- Fix carriage return in subject of email, which may block the email
- Fix crash when validating several steps at the same time in the mass update interface
- Forward port PR #236 and #240

PR #270, opened 2023-02-28
[FIX][15.0] 2 small fixes in account_invoice_overdue_reminder

alexis-via:
- remove carriage return in mail subject, which can block email generation
- Fix crash when validating several steps in mass update interface


alexis-via:
These 2 fixes are backported from #269

alexis-via:
These 2 bugs are not present in v14

PR #271, opened 2023-03-16
[15.0][FIX] partner_risk_insurance: warning message for invalid field parameter tracking

sergio-teruel:
cc @Tecnativa

This module has not dependency from mail, so the tracking parameter is invalid.

ping @carlosdauden @chienandalu

### DIFF
--- a/partner_risk_insurance/__manifest__.py
+++ b/partner_risk_insurance/__manifest__.py
@@ -12,7 +12,7 @@
     "license": "AGPL-3",
     "application": False,
     "website": "https://github.com/OCA/credit-control",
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

chienandalu (approved):

pedrobaeza (changes requested):
I'm afraid the solution is not valid, as it's tricking the system and can generate side effects. Please change the dependency instead.

pedrobaeza (approved):
Dependency changed for avoiding the warning

/ocabot merge patch

PR #281, opened 2023-04-27
[16.0] account_invoice_overdue_reminder: allow to remove invoices in the wizard

alexis-via:
Allow to remove invoices in the wizard : same feature as https://github.com/OCA/credit-control/pull/170 but the a different implementation that doesn't need a button to refresh the wizard
Remove the end step in the wizard (replaced by a notification)
Remove the up_to_date boolean on the start wizard
Code cleanup

### DIFF
--- a/account_invoice_overdue_reminder/models/account_move.py
+++ b/account_invoice_overdue_reminder/models/account_move.py
@@ -55,23 +55,14 @@ class AccountMove(models.Model):
     @api.depends(
         "overdue_reminder_ids.action_id.date",
         "overdue_reminder_ids.counter",
-        "overdue_reminder_ids.action_id.reminder_type",
     )
     def _compute_overdue_reminder(self):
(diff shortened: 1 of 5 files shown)


clementmbr (approved):
LGTM, less code and more functionalities!

alexis-via:
@legalsylvain This is the feature you added in v12, but with a different implementation (without the button to recompute the email message)

legalsylvain:
> @legalsylvain This is the feature you added in v12, but with a different implementation (without the button to recompute the email message)

I took a quick look, but did'nt understood all the code. I have to make some tests with runboat. 

That's a great news, because my code was very well designed, and with bad UX, so great if the "refresh" button is removed ! :+1:

legalsylvain:
> I took a quick look, but did'nt understood all the code. I have to make some tests with runboat.

Too late for the tests ! 

I'll take a look when I migrate in V16. 

Regards.

PR #305, opened 2023-08-16
[16.0][MIG] partner_risk_insurance: Migration to 16.0

victoralmau:
Migration to 16.0

Changes done:
- [x] The compute is removed from the `credit_limit` field to avoid inconsistencies with the same field that adds account.
- [x] Specific inverse is added for the `company_credit_limit` + `insurance_credit_limit` fields.
- [x] The value of `credit_limit` is defined with `_inverse_oca_oca_credit_limit()` to maintain the functionality of this module and that of `account`.

Please @CarlosRoca13 and @stefan-tecnativa can you review it?

@Tecnativa TT44335


partner_risk_insurance/tests/test_partner_risk_insurance.py
```diff
@@ -0,0 +1,20 @@
...
+        cls.partner = cls.env["res.partner"].create(
+            {
+                "name": "Test",
+                "company_credit_limit": 100.00,
+                "insurance_credit_limit": 50.00,
+            }
+        )
+
+    def test_credit_limit(self):
+        self.assertEqual(self.partner.credit_limit, 150.00)
```
stefan-tecnativa, line 24: It seems that the compute is not being triggered, so the test is failing.

victoralmau: Actually the problem here is that the `credit_limit` field (https://github.com/odoo/odoo/blob/16.0/addons/account/models/partner.py#L550) and its own logic are added in `account`. I am not very clear what would be the step to follow in this case (deform the standard operation, rename the field or others).

CarlosRoca13 (approved):
Code review

pedrobaeza:
Not sure about the feature of the credit limit introduced in this module, so calling @carlosdauden for his blessing.

stefan-tecnativa (approved):

rafaelbn:
@carlosdauden @Shide a last review here? This is going to be migrated

partner_risk_insurance/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+        help="Credit limit granted by the company.",
+        tracking=True,
+        inverse="_inverse_oca_credit_limit",
+    )
+    insurance_credit_limit = fields.Float(
+        "Insurance's Credit Limit",
+        help="Credit limit granted by the insurance company.",
+        tracking=True,
+        inverse="_inverse_oca_credit_limit",
+    )
```
Shide, line 33: Why not remove this field and let only the `credit_limit`? Is any difference between `insurance_credit_limit` and `credit_limit`?
On the view > insurance page, just use the `credit_limit` field with a different string (like in **account_financial_risk**) and remove the confusing `_inverse_oca_credit_limit` function and this field.

partner_risk_insurance/tests/test_partner_risk_insurance.py
victoralmau: Ping @carlosdauden

partner_risk_insurance/models/res_partner.py
rafaelbn: @victoralmau @carlosdauden we can merge as is and we can propose a PR later. Let us know please! 👍🏼

carlosdauden: A partner can be insured for a risk of 1000, but the company can assume the risk of an additional 500, so the credit limit for that partner would be 1500.

The credit_limit field already existed in previous versions, what this module did in the previous version was to change it to computed (to act as an "advanced" onchange).
In this version the field has been changed to company_depend (property), so we will have to adapt the module to this change without losing the current functionality.

victoralmau: Ok, I understand what you are trying to do, but there is also a small detail, in v15 the field was added by `base`, and in v16 the field (`credit_limit`) is added by `account`.

Taking into account that this module does not depend on `account` because `account_credit_control` already exists, what would be the process to follow in this case (maybe in this module leave the current behavior of compute, and in `account_credit_control` make it compatible with company_dependent as you indicate).

pedrobaeza: `account_credit_control` module is for other things. I think this one can depend on `account` directly.

victoralmau:
Changes done, can you check it again?

josep-tecnativa:
Please, add this changes according to the comment https://github.com/OCA/credit-control/commit/bc7a7e67468d841a4edc920634f2460192c20340#r130488985

stefan-tecnativa:
@victoralmau I've forward ported some commits from this PR: #337 Can you check that everything is okay?
ping @carlosdauden

pedrobaeza (approved):
/ocabot merge nobump

PR #306, opened 2023-08-17
[15.0][FIX] account_financial_risk: Make tests resilient

CarlosRoca13:
- Depending on the installed set of modules, the company currency may be USD or EUR. If the second case, these tests will fail, so we make sure that the company currency is USD for our tests, doing the change by SQL, as there's a Python constraint that prevents it.

Not needed in v17 due to: odoo/odoo#107113.

cc @Tecnativa


CarlosRoca13:
@pedrobaeza @victoralmau

pedrobaeza (commented):
Please reduce the diff to the minimum. I see that you are only making sure that the USD is active. Is that a real problem?

pedrobaeza (approved):

CarlosRoca13:
> Please reduce the diff to the minimum. I see that you are only making sure that the USD is active. Is that a real problem?

Yes, it can be disabled on some instances, so we need to make sure that it is active

PR #322, opened 2023-09-25
[16.0][MIG] account_payment_return_financial_risk: Migration to 16.0

victoralmau:
Migration to 16.0

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT44310


account_payment_return_financial_risk/tests/test_payment_return_risk.py
```diff
@@ -30,14 +27,19 @@ def setUpClass(cls):
...
-                "user_type_id": cls.env["account.account.type"]
-                .create(
-                    {"name": "Test income", "type": "other", "internal_group": "income"}
-                )
-                .id,
+                "account_type": "income_other",
             }
         )
         cls.partner = cls.env["res.partner"].create({"name": "Test"})
+        cls.tax = cls.env["account.tax"].create(
```
pedrobaeza: Why do you need a tax?

victoralmau: A 0% rate is used to prevent Tax 15% from self-defining and the amounts are not as expected.

carolinafernandez-tecnativa: @victoralmau If you force tax_ids = False? still happens the same?

pedrobaeza: Then just empty the field, instead of creating a new tax.

carolinafernandez-tecnativa (approved):
TT44310

pedrobaeza (approved):
/ocabot merge nobump

Issue #323, opened 2023-09-28
account_financial_risk, +related modules - Make stored all fields related to risk information


Shide:
# Context
I've reviewed all **Risk Management** (only Risk Management, not Credit Control or Reminders) modules and a lot of partner fields has a compute function with an `@api.depends`. This means that are being recomputed every time the fields in the `@api.depends` changes. The fields:

**Risk Amounts:**
- risk_invoice_draft
- risk_invoice_open
- risk_invoice_unpaid
- risk_account_amount
- risk_account_amount_unpaid

**Exceptions:**
- risk_total
- risk_exception
- risk_amount_exceeded

**Currency:**
- risk_currency_id

**Remaining Amounts:**
- risk_remaining_value
- risk_remaining_percentage

# The problem
These fields are not stored on the database and every recomputation, a precious time is wasted or not efficiently used. 
*Usually, if a field has an `@api.depends`, we store in the database to not trigger recomputation every time the field is read.*

# What I would like to change
I suggest make these fields stored on the database (the only thing needs to change is the `store=True` on each field, an easy solution) and have these fields available for search, compare, print on reports or views, etc. without an extra cost.

# Desirable option
I prefer to store in the database and have all the data stored and updated at any point.
Maybe you want to make a spreadsheet with these data, send it to another system, show a full report and having this stored in the database will help to increase efficency.

If we are agree with this change I will make all the necessary PRs for v15+ for every module that depends on `account_financial_risk` (account_payment_return_financial_risk, sale_financial_risk, sale_payment_sheet_financial_risk)

# Comments
What do you guys think about this? @rafaelbn @pedrobaeza @carlosdauden @MiquelRForgeFlow  @fclementic2c

carlosdauden:
Originally these fields were stored, but we realised that storing such ephemeral data was of no use, as the performance for calculating one client's risk is very good, and caused crashes by constantly writing the res_partner table.
These fields are designed to be consulted for one or a few customers, for example (customer form view, sales order form view, when confirming one or more sales orders, etc.).

If a specific customer needs to work massively with this data, it is very easy to add the option store=True in your custom code.

For these reasons I don't think it's a good idea to go back to a less optimal code for most uses.

PR #326, opened 2023-09-28
[15.0][FIX] sale_financial_risk: Make stored computed risk fields

Shide:
Changes related to https://github.com/OCA/credit-control/issues/323

Change PR series: 
https://github.com/OCA/credit-control/pull/324
https://github.com/OCA/credit-control/pull/325
https://github.com/OCA/credit-control/pull/326
https://github.com/OCA/credit-control/pull/327

MT-3681 @moduon

### DIFF
--- a/sale_financial_risk/models/res_partner.py
+++ b/sale_financial_risk/models/res_partner.py
@@ -17,6 +17,8 @@ class ResPartner(models.Model):
     )
     risk_sale_order = fields.Monetary(
         compute="_compute_risk_sale_order",
+        compute_sudo=True,
+        store=True,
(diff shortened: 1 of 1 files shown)


carlosdauden:
Originally these fields were stored, but we realised that storing such ephemeral data was of no use, as the performance for calculating a client's risk is very good, and caused crashes by constantly writing the res_partner table.
These fields are designed to be consulted for one or a few customers, for example (customer form view, sales order form view, when confirming one or more sales orders, etc.).

If a specific customer needs to work massively with this data, it is very easy to add the option store=True in your custom code.

For these reasons I don't think it's a good idea to go back to a less optimal code for most uses.

PR #335, opened 2023-10-19
[16.0][IMP] account_financial_risk: fix partner view

OriolMForgeFlow:


account_financial_risk/static/src/scss/account_financial_risk_table.scss
```diff
@@ -6,3 +6,7 @@
         text-align: right;
     }
 }
+
+.oe_risk_force_margin_left {
+    margin-left: 0 !important;
```
pedrobaeza, line 11: Can you avoid the use of `!important`

pedrobaeza (commented):
Please put the before and after screenshots

OriolMForgeFlow:
Yeah, sure. Here's the before:
image


And here's the after:
image

account_financial_risk/static/src/scss/account_financial_risk_table.scss
OriolMForgeFlow: If the `!important` rule is not set, it won't take it into account. Instead, it will use the margin-left: auto property from the oe_subtotal_footer class.

pedrobaeza:
But it should be a way to do this without CSS, isn't it?

pedrobaeza:
The question is that I see too many changes for only avoiding an extra underline.

MiquelRForgeFlow:
> I see too many changes

This is due to xml alignment.

MiquelRForgeFlow (approved):

pedrobaeza:
OK, but I still think that this should be fixed without CSS tricks and `!important` one. Please do it the same as in the left part.

PR #337, opened 2023-10-23
[15.0][IMP] partner_risk_insurance: Add invoice insurance risk report

stefan-tecnativa:
TT38748 @Tecnativa

@carlosdauden please review!

This PR adds a custom qweb report for invoices published but pending of payment, sorted by the conditions of whether the customer is insured or not, the country and by their name. For now, it's considered an insured customer the one who has it's `risk_insurance_grant_date` field assigned to a date.


partner_risk_insurance/wizard/invoice_risk_insurance_wizard.py
```diff
@@ -0,0 +1,46 @@
...
+
+from odoo import _, fields, models
+from odoo.exceptions import ValidationError
+
+
+class InvoiceRiskInsuranceWizard(models.TransientModel):
+    _name = "partner.risk.insurance.wizard"
+    _description = "Wizard to see invoices sorted by risk"
+
+    date_since = fields.Date(
```
carlosdauden: Although the term used is correct, we usually use date_from and date_to to name these fields.
They also match better with the strings used.

partner_risk_insurance/wizard/invoice_risk_insurance_wizard.py
```diff
@@ -0,0 +1,46 @@
...
+                )
+            )
+        moves = moves.sorted(
+            lambda m: (
+                m.partner_id.risk_insurance_grant_date and 1 or 0,
+                m.partner_id.country_id,
+                m.name,
+            )
+        )
+        # import wdb;wdb.set_trace()
```
carlosdauden: Remove

partner_risk_insurance/templates/invoice_risk_insurance_template.xml
```diff
@@ -0,0 +1,170 @@
...
+                </strong>
+            </td>
+        </tr>
+    </template>
+    <template id="report_invoice_risk_insurance_document">
+        <t t-call="web.external_layout">
+
+            <t
+                t-set="insurance_total_amount"
+                t-value="sum(docs.mapped('amount_total'))"
```
carlosdauden: Why do you need to go through all the records again to calculate the total instead of accumulating it in the same way as you accumulate the total per country?

partner_risk_insurance/templates/invoice_risk_insurance_template.xml
```diff
@@ -0,0 +1,170 @@
...
+            />
+            <t t-set="change_insurance_type" t-value="False" />
+            <t
+                t-set="insured_moves"
+                t-value="docs.filtered(lambda mv: mv.partner_id.risk_insurance_grant_date)"
```
carlosdauden: Can't the same strategy be used as for countries instead of going through all the records several times to filter them?

carlosdauden (changes requested):

partner_risk_insurance/templates/invoice_risk_insurance_template.xml
stefan-tecnativa: The problem is that I can't accumulate the total because it's value is not stored between the two templates that I am using to render the report. I changed how is the partial calculated, but as I said for now I don't know how this value could be preserved.

partner_risk_insurance/templates/invoice_risk_insurance_template.xml
stefan-tecnativa: This way makes the code more understandable, but it is true that this consultations to record are unnecessary. It is prefearable to follow the same strategy for the entire report?

carlosdauden (approved):

chienandalu (approved):

pedrobaeza:
Please add it for v16 in https://github.com/OCA/credit-control/pull/305.

/ocabot merge major

PR #341, opened 2023-11-22
[15.0][FIX] account_financial_risk: Recover original behavior of include checks to be not need to compute risk exception

sergio-teruel:
cc @Tecnativa TT46129

ping @carlosdauden @victoralmau

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -433,7 +433,7 @@ class ResPartner(models.Model):
                 field_value = getattr(partner, risk_field[0], 0.0)
                 max_value = getattr(partner, risk_field[1], 0.0)
                 include = getattr(partner, risk_field[2], False)
-                if include and max_value and field_value > max_value:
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

carlosdauden (approved):

carlosdauden:
@MiquelRForgeFlow What was the business need for this change, any problems with decimals?

https://github.com/OCA/credit-control/pull/161/commits/3147161ed2ee8bbf58188ae6c6ce94cd5f9884b0#diff-fc5d87b63ee69d5438b5a182e69fa68d82de8ef90e8d03e2279ab1057f7dc19dR384

MiquelRForgeFlow:
I don't remember. But maybe it was a client demand. If you could add a hook in that line to enable possible customization, it would be nice.

pedrobaeza:
Let's fix it, and anyone can add the hook in any moment:

/ocabot merge patch

PR #342, opened 2023-11-24
[16.0][FIX] account_financial_risk: Recover original behavior of include checks to be not need to compute risk exception

pedrobaeza:
Forward-port of #341 

@Tecnativa


pedrobaeza:
Fast-tracking as already merged in previous version for closing the cycle:

/ocabot merge patch

PR #347, opened 2023-12-14
[17.0][MIG] account_financial_risk: Migration to 17.0

Mat-moran:
- Change version in manifest
- Adapt views to version 17.0


Mat-moran:
Hi, @pedrobaeza can you help on this? we have a  problem with the field `amount_residual_currency` when the method`_get_amount_in_risk_currency()` is executed.

The field `amount_residual_currency` when is defined it explicetily as
```python
amount_residual_currency = fields.Monetary(
    string='Residual Amount in Currency',
    compute='_compute_amount_residual', store=True,
    group_operator=None, # This is making the function fail
    help="The residual amount on a journal item expressed in its currency (possibly not the "
         "company currency).",
)
```

what do you sugest?

pedrobaeza:
Sorry, I can't say, as I haven't dug on it.

account_financial_risk/data/account_financial_risk_data.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+        <!-- Account Tax Group -->
+        <record id="tax_group_taxes" model="account.tax.group">
```
carlosdauden, line 4: Why this module adds new tax group?

account_financial_risk/models/res_partner.py
```diff
@@ -354,7 +354,6 @@ def _compute_risk_account_amount(self):
                 group["domain"] + [("partner_id", "in", customers.ids)],
                 group["fields"],
                 group["group_by"],
-                orderby="id",
```
carlosdauden: Use _read_group to remove this

account_financial_risk/data/account_financial_risk_data.xml
carlosdauden, line 4: Why this module adds new tax group?

carlosdauden (changes requested):
Ignore not related files as macOS.pkg

Take a look at:
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#7git

account_financial_risk/data/account_financial_risk_data.xml
Mat-moran: Thanks for the review,

This was in odoo account module but has been removed in v17. As it is required by the test, so we added to this module.

```xml
# before v17 this data was on -> addons/account/data/account_data.xml
<!-- Account Tax Group -->
<record id="tax_group_taxes" model="account.tax.group">
    <field name="name">Taxes</field>
    <field name="sequence">0</field>
</record>
```

account_financial_risk/models/res_partner.py
Mat-moran: I am not sure if I am understanding. 

the return of `_read_group` is not the same as `read_group`

ioans73:
@Mat-moran what is the status of this?

account_financial_risk/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2016-2018 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Account Financial Risk",
+    "summary": "Manage customer risk",
+    "version": "17.0.1.0",
```
sbiosca-s73, line 7: Please, put the version in the correctly form 
`17.0.1.0.0`

pedrobaeza, line 7:
```suggestion
    "version": "17.0.1.0.0",
```

account_financial_risk/data/account_financial_risk_data.xml
pedrobaeza: You should fix the test, not to add useless data.

account_financial_risk/models/res_partner.py
carlosdauden: Look this:
https://github.com/OCA/credit-control/pull/377#pullrequestreview-2035518788

pedrobaeza:
Superseded by #379

PR #352, opened 2024-01-03
[15.0][IMP] partner_risk_insurance: Add only unpaid option to invoice selection

carlosdauden:
Add only unpaid option to allow include/exclude invoices in payment state not paid

@Tecnativa TT46889

### DIFF
--- a/partner_risk_insurance/wizard/invoice_risk_insurance_wizard.py
+++ b/partner_risk_insurance/wizard/invoice_risk_insurance_wizard.py
@@ -13,17 +13,23 @@ class InvoiceRiskInsuranceWizard(models.TransientModel):
         string="From", required=True, default=fields.Date.context_today
     )
     date_to = fields.Date(string="To", required=True, default=fields.Date.context_today)
+    only_unpaid = fields.Boolean(
+        help="If check, includes only invoices in state not paid"
+    )
+
+    def get_moves_domain(self):
+        domain = [
+            ("state", "=", "posted"),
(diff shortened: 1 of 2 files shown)


partner_risk_insurance/wizard/invoice_risk_insurance_wizard.py
```diff
@@ -13,31 +13,30 @@ class InvoiceRiskInsuranceWizard(models.TransientModel):
...
-        )
+        moves = self.env["account.move"].search(self.get_moves_domain())
         if not moves:
             raise ValidationError(
                 _(
                     "There aren't moves between this dates\n"
                     "Please, select different dates"
                 )
             )
-        moves = moves.sorted(
```
pedrobaeza, line 34: Isn't this sort not needed anymore?

sergio-teruel (approved):

pedrobaeza (approved):

partner_risk_insurance/wizard/invoice_risk_insurance_wizard.py
carlosdauden: Sorted in template because seems that the order is lost in the report_action call:
https://github.com/OCA/credit-control/blob/39119989c54e1198d17403372f032170604b1ce9/partner_risk_insurance/templates/invoice_risk_insurance_template.xml#L21-L22

pedrobaeza:
Please don't forget to fw-port all of this to v16.

PR #354, opened 2024-01-22
[16.0][IMP] account_invoice_overdue_reminder: add 3 read-only ACLs for auditor group

alexis-via:


rafaelbn (approved):
@alexis-via I guess this could be merged?

PR #357, opened 2024-02-12
[15.0][MIG] account_invoice_overdue_warn_sale: Migration to 15.0

siemenv:
Depends on:

- https://github.com/OCA/credit-control/pull/356


siemenv:
This module depends on "account_invoice_overdue_warn" which is also in PR.

PR #359, opened 2024-02-13
[15.0][IMP] partner_risk_insurance: Performance installation on large databases

sergio-teruel:
cc @Tecnativa 

ping @carlosdauden @Shide @rafaelbn

PR #355, opened 2024-02-02
[15.0] [IMP] partner_risk_insurance: Insure invoices with credit policies

Shide:
This improvement allows to insure invoices with credit policies that has marked as "Insure invoices".

When a partner is set on an invoice (only out_invoice) and has a credit policy marked as "Insure invoices", the invoice will be marked as "Insured by a Credit Policy" and the policy and company policy will be set on the invoice.

You can filter on the tree view by the "Insured by a Credit Policy" and group by "credit_policy_company_id" out of the box.

Refactors:
- Added original creators on the top of some files
- Added group (account.group_account_user) to the Insurance page on Partner.
- Added myself as a maintainer too
- Renamed credit_policy.py to credit_policy_state.py because there is another related model to credit policies in models folder. (Sorry, only added active and insure_invoices fields)

Roadmap: 
- Be able to print an image on the Insured invoices (this is why i add the the credit.policy.company model)
- FWP this and next PRs to v16

Small demo:
https://www.loom.com/share/ac1ce65877644820a5ed53ca08fc38cc?sid=0cae7ed3-de3c-45a3-ac62-949457a49ec4

MT-4974 @moduon @rafaelbn @fcvalgar @yajo @edlopen @sergio-teruel please review if you want :)

PR #360, opened 2024-02-15
[16.0] [FWP] partner_risk_insurance: PR 355, PR 359

Shide:
FWP of PRs:
- https://github.com/OCA/credit-control/pull/355
- https://github.com/OCA/credit-control/pull/359
- Added commit to fix tests on `account_payment_return_financial_risk`

MT-4974 MT-4827 @moduon @rafaelbn @sergio-teruel @carlosdauden @pedrobaeza please review if you want :)


yajo (changes requested):
It seems this change makes `account_payment_return_financial_risk` fail... could you check tests please?

partner_risk_insurance/models/account_move.py
```diff
@@ -36,14 +36,15 @@ class AccountMove(models.Model):
     @api.depends("partner_id", "move_type")
     def _compute_insured_with_credit_policy(self):
         """Compute the insured_with_credit_policy field when partner changes."""
-        self.update(
+        self_w_ctx = self.with_context(check_move_validity=False)
```
carlosdauden, line 39: ¿Why change ths instead of add `check_move_validity` in the next dictionary?

carlosdauden (changes requested):

partner_risk_insurance/models/account_move.py
Shide: Without this context, [account_payment_return_financial_risk](https://github.com/OCA/credit-control/tree/16.0/account_payment_return_financial_risk) module fails when running tests.

`account_payment_return_financial_risk` depends on [account_payment_return](https://github.com/OCA/account-payment/tree/16.0/account_payment_return) that creates move lines [one by one](https://github.com/OCA/account-payment/blob/16.0/account_payment_return/models/payment_return.py#L185-L187) (whitout the counterpart) and making a trigger on `_compute_insured_with_credit_policy`, causing the `self.update(...)` fail because the move is not balanced when adding the lines.

Since this module is modifying the `account.move` itself and not the lines nor amounts, I've considered safe to skip move validations and let the other module tests run without issues.

carlosdauden (approved):

rafaelbn (approved):
/ocabot merge major

PR #362, opened 2024-02-21
[IMP] partner_risk_insurance: improve UX

yajo:
In Odoo 16, upstream has already a `credit_limit` field in `res.partner`. That field is coupled with a checkbox named `use_partner_credit_limit`, which means "this partner doesn't use the default credit limit; it has a specific credit limit".

Those fields were only available to users of groups `account.group_account_invoice` and `account.group_account_readonly`, within the *Invoicing* tab in `res.partner` form.

image

Still there's an extra hidden `show_credit_limit` field which toggles this feature completely. It depends on wether the current company has credit limits enabled.

This module was migrated from previous versions that had none of these features. The result is that, probably by that inertia, UX ended up becoming inconsistent.

Improvements done:
- Fields are relabeled. Their names confused everybody in our UX investigation. They now reflect better their usage.
- The company default credit limit is ignored when the user sets a different credit limit. Before, it was being included in the sum to get the total credit limit for the partner, but that was contrary to the purpose of that field in upstream Odoo. When you check `use_partner_credit_limit` in upstream Odoo, you don't add an extra credit limit; instead, you *replace* the default limit by a custom one. That's what we do here now.
- Tests are rewritten because they were exercising that wrong assumption that I just explained.
- It made no sense to ask the user for total, company and insured credit limits. One of those fields could be computed. Since upstream Odoo already has a writable property field `credit_limit` (already explained), I picked up `company_credit_limit` to be the computed one.
- `company_credit_limit` is no longer stored (and therefore not tracked) because it's a company-dependant field. In multi-company partners, the logger became inconsistent otherwise.
- `use_partner_credit_limit` gets an onchange to trigger its inverse method beforehand, and make UX consistent before hitting save.
- When setting a insurance credit limit higher than the total credit limit, the total is automatically increased, and the company is automatically set to 0.
- All features are hidden if the company didn't enable credit limits. This is consistent with upstream Odoo.
- Permissions are adjusted to match those of upstream Odoo regarding these fields.
- Tests are executed with those new permissions.

Still, a good extra enhancement would be to move all this stuff to the *Invoicing* tab. But that can be done later after some more investigation.

@moduon MT-4827


rafaelbn:
Hello @carlosdauden ,

We are trying to improve UX.

Your point of view is really welcome!

Thanks

rafaelbn (commented):
Here you are Jairo the labels and descriptions. More standard terminology.

| English Field Name      | Spanish Field Name           |
|-------------------------|------------------------------|
| Company Credit Limit    | Límite de Crédito de la Empresa |
| Insured Credit Limit    | Límite de Crédito Asegurado  |
| Total Credit Limit      | Límite de Crédito Total      |

- **Company Credit Limit**: **Calculated** as the Total Credit Limit minus the Insured Credit Limit, reflecting the credit limit the company is willing to offer to the client, excluding the insured portion.
- **Insured Credit Limit**: The portion of credit that is covered by insurance, **set by the user**. It represents the amount of credit that an insurance policy guarantees.
- **Total Credit Limit**: The overall credit limit available to a client, **set by the user**. It includes both the company's own credit limit and the insured credit limit.

Thank you!

rafaelbn (approved):
I agree setting computed the **company_credit_limit**. Rationale in 1 minute:

https://www.loom.com/share/65a27359c58d4147bbada6fa4938c71e?sid=f7fc2260-8000-42f4-8d23-a1116b77c981


I propose in [other comment](https://github.com/OCA/credit-control/pull/362#pullrequestreview-1894161819) the names of the labels and explanations of the field, even though your are also valid and explicit, then only one I would change, al least, is Credit Limit to **Total Credit Limit**

- [x] Change is Credit Limit to **Total Credit Limit**


@flachica @victoralmau @carlosdauden do you agree rationale?

Shide (approved):
**praise:** Super cool UX improvement! ❤️ 
**todo:** Fix tests

Code and functional review exporting and importing data

https://www.loom.com/share/5faeab438fb547cd85f6f8323b5d8812?sid=b1665b87-c1e1-4dae-b03e-aaf84f027fe0

yajo:
I pushed a fix to `account_credit_control` that should fix the builds.

yajo:
All relabeling suggestions applied.

Gelojr (approved):
Very nice job. Thanks @yajo I love it!!
Attached is a test video with several use cases and it works propertly.
https://www.loom.com/share/e04a41e1ac4e4e078e11b9a3008ded38?sid=2437d817-45f8-4e51-b2b4-3c029f27b07c

carlosdauden (approved):

yajo:
Can someone merge please?

rafaelbn (approved):
/ocabot merge major

PR #364, opened 2024-02-22
[15.0][FIX] sale_financial_risk: fix test

pilarvargas-tecnativa:
The test_manual_currency_risk_not_exceeded test that assesses that the manual currency risk has not been passed fails when another module changes the company's currency. The reason for the failure is that if the company currency matches the currency of the sell order it does not convert correctly and in that case it interprets the limit within the same currency. For example, if the company currency is set as EUR and the sales order is placed in EUR, the conversion of the 100€ is not done and therefore exceeds the limit set to 99. To solve this, the company currency is first checked to set a different currency in the order. On the other hand, the tests should be frozen at a certain date to prevent the exchange rate of the exchange currency from going up or down thus avoiding errors when running the test, which should check that it has not been exceeded.

```
2024-02-22 00:00:00,000 1 ERROR devel odoo.addons.sale_financial_risk.tests.test_partner_sale_risk: FAIL: TestPartnerSaleRisk.test_manual_currency_risk_not_exceeded
Traceback (most recent call last):
  File "/opt/odoo/auto/addons/sale_financial_risk/tests/test_partner_sale_risk.py", line 224, in test_manual_currency_risk_not_exceeded
    self.assertEqual(result, True)
AssertionError: {'type': 'ir.actions.act_window', 'name':[114 chars]new'} != True
```

cc @Tecnativa

@pedrobaeza @carolinafernandez-tecnativa please review

### DIFF
--- a/sale_financial_risk/tests/test_partner_sale_risk.py
+++ b/sale_financial_risk/tests/test_partner_sale_risk.py
@@ -32,9 +32,7 @@ class TestPartnerSaleRisk(TransactionCase):
         )
         cls.main_currency = cls.env.company.currency_id
         cls.EUR = cls.env.ref("base.EUR")
-        cls.other_company = cls.env["res.company"].create(
-            {"name": "Company 2", "currency_id": cls.EUR.id}
-        )
+        cls.USD = cls.env.ref("base.USD")
         cls.sale_order = cls.create_sale_order(cls.main_currency, cls.env.company)
         cls.env.user.lang = "en_US"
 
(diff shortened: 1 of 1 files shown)


carolinafernandez-tecnativa (approved):
LGTM :) Thanks, please forward to v16

pedrobaeza:
Please indicate all the information here and in the commit message:

- Which test was failing
- Why this solves it

I wonder if the problem is that the currency rate is changed and it was 100 in other currency, but the result in the company currency is less than 99.

pedrobaeza (approved):
/ocabot merge patch

PR #380, opened 2024-05-23
[16.0][FIX] account_credit_control: replace final due amount by balanced amount

StephaneMangin:
As an invoice can be partially paid, the total due amount should reflect this state.
Changing the communication summary amounts by correct ones.

### DIFF
--- a/account_credit_control/models/credit_control_communication.py
+++ b/account_credit_control/models/credit_control_communication.py
@@ -203,11 +203,11 @@ class CreditControlCommunication(models.Model):
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
-                format_amount(self.env, line.invoice_id.amount_total, line.currency_id),
+                format_amount(self.env, line.amount_due, line.currency_id),
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
-                format_amount(self.env, line.amount_due, line.currency_id),
+                format_amount(self.env, line.balance_due, line.currency_id),
             )
             table_content += "<tr>%s</tr>" % tr_content
         table_content += "</table>"


cyrilmanuel (approved):

leemannd (changes requested):
You're changing the behavior of the module. It can be a valid change, but if you're doing so please update also the headers of the table to reflect this change:
- https://github.com/OCA/credit-control/pull/380/files#diff-ab2ca28045e4ca81ddb0cacf900797263a04da3c88c8f3f46b93ebeca3b2d9bbL184
-  https://github.com/OCA/credit-control/pull/380/files#diff-ab2ca28045e4ca81ddb0cacf900797263a04da3c88c8f3f46b93ebeca3b2d9bbL185

StephaneMangin:
> You're changing the behavior of the module. It can be a valid change, but if you're doing so please update also the headers of the table to reflect this change:
> 
> * https://github.com/OCA/credit-control/pull/380/files#diff-ab2ca28045e4ca81ddb0cacf900797263a04da3c88c8f3f46b93ebeca3b2d9bbL184
> * https://github.com/OCA/credit-control/pull/380/files#diff-ab2ca28045e4ca81ddb0cacf900797263a04da3c88c8f3f46b93ebeca3b2d9bbL185

It is the same concept, invoice total and balanced total (actual amount due). Fields naming is technical, not the title of the header which is functional

imlopes (approved):
LGTM

pedrobaeza (changes requested):
The invoice total shouldn't be changed. It's reflecting the total of the invoice, no matter the amount due, and I don't understand the other change. Why `amount_due` is not valid and you are putting `balance_due`?

StephaneMangin:
> The invoice total shouldn't be changed. It's reflecting the total of the invoice, no matter the amount due, and I don't understand the other change. Why `amount_due` is not valid and you are putting `balance_due`?

@pedrobaeza 
Hi Pedro, in fact `amount_due` is the invoice total amount (`invoice_id.amount_currency`) and `balance_due` is the real actual due amount in the particular context of a partial invoice (`invoice_id.amount_residual_currency`).

When `amount_due` is used in the summary, it is wrong with a residual due amount.

Or i'm wrong understanding this implementation. Let me know

pedrobaeza:
The total of the invoice should be reflected. Example: Invoice: 1000 €. Due: 500 €. I don't know the difference between `amount_due` and `balance_due` though.

StephaneMangin:
> 1000 €. Due: 500 €. I don't know the difference between `amount_due` and `balance_due` though.

I just explained this difference in my previous com. Maybe not so clearly though.

From what I can see right now on this piece of code (and after a client bugfix request): 
 - amount due is the invoice/move line total
 - balance_due is the residual amount for a partial invoice

When i use these two fields in the summary values are valid and reflect what to pay.

PR #381, opened 2024-05-23
[16.0][IMP] sale_financial_risk_info: groups to risk_info field

lmchlfl:
A user without assigned risk permissions can view the 'Risk Info' field.

image
image

Without having User or Manager permissions, you should not see that field.

### DIFF
--- a/sale_financial_risk_info/views/sale_order_view.xml
+++ b/sale_financial_risk_info/views/sale_order_view.xml
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-
     <record id="view_order_form" model="ir.ui.view">
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='validity_date']" position="before">
-                <field name="risk_info" readonly="1" />
+                <field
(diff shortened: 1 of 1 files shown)


flalexg (approved):
LGTM!

carlosdauden (changes requested):
Well, the objective of this module is for any salesperson with the minimum permissions to see this information, but not have access to the customer risk info tab.

If you want, you can add a new view with a configuration option that activates one or the other, but without changing the default behavior of the module.

carlosdauden:
You also have the option to customize the system parameter sale_financial_risk_info.info_pattern so that, for example, it only shows the percentage

lmchlfl:
> You also have the option to customize the system parameter sale_financial_risk_info.info_pattern so that, for example, it only shows the percentage

I've searched that pattern and it doesn't exist

carlosdauden:
Because we use a default value, but you can create this parameter sale_financial_risk_info.info_pattern to be used here:
https://github.com/OCA/credit-control/blob/52e46b39e91486630944578261a27a6debbbb526/sale_financial_risk_info/models/sale.py#L20-L24

carlosdauden:
This is explained in readme:
https://github.com/OCA/credit-control/blob/16.0/sale_financial_risk_info/README.rst#configuration

PR #392, opened 2024-07-24
[17.0][MIG] account_payment_return_financial_risk: Migration to version 17.0

carlos-lopez-tecnativa:
@Tecnativa TT49282

Depends on:

- https://github.com/OCA/account-payment/pull/752


carlosdauden:
Can you add account_payment_return repo dependency to tray in runboat?

carlos-lopez-tecnativa:
> Can you add account_payment_return repo dependency to tray in runboat?

@carlosdauden done, could you please update review

carlosdauden (approved):

carlos-lopez-tecnativa:
@carolinafernandez-tecnativa  @pedrobaeza could you please help me with the review of this module? Thank you!

account_payment_return_financial_risk/readme/USAGE.md
```diff
@@ -0,0 +1,10 @@
+To use this module, you need to:
+
+1.  Go to *Customers \> Financial Risk*
+2.  Set limits and choose options to compute in credit limit.
+3.  Create an invoice and pay it.
+4.  Create a payment return.
+5.  Go to *Invoicing/Accounting \> Customers \> Invoices* and create new
+    customer invoices.
+6.  Test the restriction trying to create an invoice for the partner for
+    an amount higher of the limit you have set.
```
carolinafernandez-tecnativa, line 10: It is missing the steps you need to set up which field in financial risk.
Maybe is needed a CONFIGURATION.md, to explain we have to add groups to user Financial/Manager to allow to edit the amounts, and which field we need to set up.

Also here is missing the step to create a customer return

account_payment_return_financial_risk/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,4 @@
+Extends Account Financial Risk to manage payments returns.
+
+If any limit is exceed the partner gets forbidden to confirm sale
```
carolinafernandez-tecnativa, line 3: This does not belong to this module description
As i understand, you wont be able to confirm a payment return if the limit is exceed.

carolinafernandez-tecnativa (commented):
I tested with this customer with this set up:

image

I created this invoice

image


I registered full payment and i go to create customer payment return for more than limit established

And i could confirm without the control

image

account_payment_return_financial_risk/readme/DESCRIPTION.md
carlos-lopez-tecnativa: @carolinafernandez-tecnativa  the purpose of this module is to include the amount of payment returns, but it is not restricted to validating a payment return. @carlosdauden, could you clarify the scope of the module?

@carolinafernandez-tecnativa  please remember that this module is just an additional extension of the main module https://github.com/OCA/credit-control/tree/17.0/account_financial_risk

image

carlos-lopez-tecnativa: Ping @carlosdauden could you please give your feedback on the above question?

account_payment_return_financial_risk/readme/USAGE.md
carlos-lopez-tecnativa: It is in the original module, so I don't think we need to add the same in this module if it is only an extension.

https://github.com/OCA/credit-control/blob/17.0/account_financial_risk/readme/CONFIGURE.md

pedrobaeza: You can add a mention saying "Refer to the module ... configuration for initial steps.".

account_payment_return_financial_risk/readme/DESCRIPTION.md
carlosdauden: Carlos you are right, what this module does is to add a new section to the financial risk to be able to manage that amount in the same way as the rest.

carlosdauden:
Ready to merge

carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #393, opened 2024-08-01
[IMP] account_financial_risk: Send email when risk confirmed

SirAionTech:
Allow to send a configurable email template when risk for an invoice is explicitly exceeded in the wizard.

Also track the credit limit date because it is just as important as the amount (that is already tracked).

### DIFF
--- a/account_financial_risk/wizards/parner_risk_exceeded.py
+++ b/account_financial_risk/wizards/parner_risk_exceeded.py
@@ -1,4 +1,5 @@
 # Copyright 2016-2018 Tecnativa - Carlos Dauden
+# Copyright 2024 Simone Rubino - Aion Tech
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, fields, models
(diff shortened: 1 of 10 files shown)


account_financial_risk/wizards/parner_risk_exceeded.py
```diff
@@ -31,8 +32,17 @@ def action_show(self):
             "target": "new",
         }
 
+    def _post_account_move_confirm_risk(self, origin_record):
+        template = origin_record.company_id.account_move_confirm_risk_template_id
+        if template:
+            template.send_mail(origin_record.id)
+
     def button_continue(self):
         self.ensure_one()
```
carlosdauden, line 46:
```suggestion
        self.ensure_one()
        res = getattr(
            self.origin_reference.with_context(bypass_risk=True), self.continue_method
        )()
        self._post_confirm_risk()
        return res
```

account_financial_risk/wizards/parner_risk_exceeded.py
```diff
@@ -31,8 +32,17 @@ def action_show(self):
             "target": "new",
         }
 
+    def _post_account_move_confirm_risk(self, origin_record):
+        template = origin_record.company_id.account_move_confirm_risk_template_id
```
carlosdauden:
```suggestion
    def _post_confirm_risk(self):
        origin_record = self.origin_reference
        model_preffix = origin_record._name.replace('.', '_')
        template = getattr(origin_record.company_id, f"{model_preffix}_confirm_risk_template_id", False)
```

account_financial_risk/wizards/parner_risk_exceeded.py
```diff
@@ -31,8 +32,17 @@ def action_show(self):
...
         self.ensure_one()
-        return getattr(
-            self.origin_reference.with_context(bypass_risk=True), self.continue_method
+        origin_record = self.origin_reference
+        continue_result = getattr(
+            origin_record.with_context(bypass_risk=True), self.continue_method
         )()
+        if origin_record._name == "account.move":
+            self._post_account_move_confirm_risk(origin_record)
+        return continue_result
```
carlosdauden, line 53:
```suggestion
```

carlosdauden (changes requested):
Try this to simplify the extension with other modules as #394

account_financial_risk/wizards/parner_risk_exceeded.py
SirAionTech: Hi, thanks for your suggestions!
Sure this would simplify code for other modules, but I think it would become too hard to understand what's happening as it sometimes happens when reflection is used.

I created this hook to explicitly do things for `account.move` records, (that's why its name is `_post_*account_move*_confirm_risk`); I've added documentation and modified its signature to make it more clear, let me know if that's ok for you.

PR #394, opened 2024-08-01
[IMP] sale_financial_risk: Send email when risk confirmed

SirAionTech:
Allow to send a configurable email template when risk for an invoice is explicitly exceeded in the wizard.

Similar to https://github.com/OCA/credit-control/pull/393, but for sale orders.


SirAionTech:
Rebased on `16.0`.

sale_financial_risk/wizards/partner_risk_exceeded.py
```diff
@@ -0,0 +1,20 @@
...
+        template = origin_record.company_id.sale_order_confirm_risk_template_id
+        if template:
+            template.send_mail(origin_record.id)
+
+    def button_continue(self):
+        continue_result = super().button_continue()
+        origin_record = self.origin_reference
+        if origin_record._name == "sale.order":
+            self._post_sale_order_confirm_risk(origin_record)
+        return continue_result
```
carlosdauden: Unnecessary with the changes proposed here https://github.com/OCA/credit-control/pull/393#pullrequestreview-2472184800

carlosdauden (changes requested):

sale_financial_risk/wizards/partner_risk_exceeded.py
SirAionTech: Thanks for having a look!
As I wrote in https://github.com/OCA/credit-control/pull/393#discussion_r1865620128, I think it's better to have the behavior for sale orders explicitly managed, so I have improved the documentation of this method to explain that.

PR #402, opened 2024-10-10
[IMP] Add a config and a hook to sort moves using partner or residual  amount

lmarion-source:
The initial behavior was only planned for the use of the residual amount as a sorting criterion.
One might need to sort by partner instead. This hook allows us to change the sorting using a config parameter


acsonefho (approved):
Code LGTM

PR #403, opened 2024-10-22
[14.0][ADD] partner_risk_insurance_security

toita86:
Currently, anyone can view and edit the Credit Insurance tab added by the module `partner_risk_insurance`.

Implementation: Created a group to restrict access to the fields added by the aforementioned module.


pedrobaeza:
This must be an extra module, as it's modifying current behavior heavily. The reason to not protect it is because not everyone can edit contacts, and those who can, needs a lot of times to touch these amounts (they are salesmen providing the information).

pedrobaeza:
A usual suffix is `_security` for this kind of things.

toita86:
> A usual suffix is `_security` for this kind of things.

Thanks for the info!

francesco-ooops (approved):
Functional ok!

partner_risk_insurance_security/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    is_full_access_user = fields.Boolean(
+        string="Is Full Access User",
+        compute="_compute_is_full_access_user",
+        store=False,
+    )
+
+    @api.depends("user_id")
```
HekkiMelody: **issue**: The field is not stored, so you don't even need to have this `@api.depends` at all, since it'll be recomputed every time anyway.

You can have the `@api.depends` if you want the field to be recomputed based on specific conditions, in addition to it being recomputed almost every time you look at the record, though.

In this case, you're asking Odoo to recompute the field based on the field `user_id` of `res.partner`, i.e. Salesman. In other words, when you change "Salesman" on a contact, recompute the value of this field.

I think what you're looking for is `@api.depends_context('uid')`

partner_risk_insurance_security/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+    def _compute_is_full_access_user(self):
+        for partner in self:
+            partner.is_full_access_user = (
+                self.env.user
+                in self.env.ref(
+                    "partner_risk_insurance_security.group_full_access_credit_policy_state"
+                ).users
+            )
```
HekkiMelody: **suggestion**: I think `self.env.user.has_group("partner_risk_insurance_security.group_full_access_credit_policy_state")` is simpler to write and read, and probably also slightly better in terms of performance.

partner_risk_insurance_security/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+
+    is_full_access_user = fields.Boolean(
+        string="Is Full Access User",
+        compute="_compute_is_full_access_user",
+        store=False,
+    )
+
+    @api.depends("user_id")
+    def _compute_is_full_access_user(self):
+        for partner in self:
```
HekkiMelody: **issue**: You're not using any field of the partner. You don't need to recompute the value of the field for every partner, you can compute it once and then assign it to all.

Even better, you can assign it all at once for every partner:

```py
self.is_full_access_user = ....
```

partner_risk_insurance_security/views/res_partner_view.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!--Add risk insurance tab in partner form. -->
+    <record model="ir.ui.view" id="view_partner_form">
+        <field name="name">res.partner.form.risk_insurance_security</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="priority" eval="50" />
```
HekkiMelody: **issue**: The fields you're modifying do not exist in `base.view_partner_form`. I think you want to inherit from `partner_risk_insurance.view_partner_form`.

**suggestion** You can also drop the `priority` IMO, if you don't need it for anything in particular.

partner_risk_insurance_security/views/res_partner_view.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!--Add risk insurance tab in partner form. -->
```
HekkiMelody: **chore**: leftover comment from copy-pasting.

partner_risk_insurance_security/views/res_partner_view.xml
```diff
@@ -0,0 +1,66 @@
...
+            </xpath>
+            <xpath expr="//field[@name='credit_policy_state_id']" position="attributes">
+                <attribute
+                    name="attrs"
+                >{'readonly': [('is_full_access_user', '=', False)]}</attribute>
+            </xpath>
+            <xpath expr="//field[@name='credit_policy_state_id']" position="after">
+                <field name="is_full_access_user" invisible="1" />
+            </xpath>
+        </field>
```
HekkiMelody, line 61: **question**: can you add `readonly` directly on the `<page name="credit_insurance position="attributes">` element instead of adding it on each field inside it?

partner_risk_insurance_security/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_full_access_credit_policy_state,res.partner,base.model_res_partner,partner_risk_insurance_security.group_full_access_credit_policy_state,1,1,1,1
```
HekkiMelody: **issue**: access to `res.partner` is already managed by other security groups, and is not strictly needed for this module to work. I'm not sure this specific groups needs to grant full CRUD access to `res.partner`.

My instinct would be to remove this file entirely.

HekkiMelody (changes requested):

HekkiMelody:
Also, not strictly needed for such a simple module IMO, but always nice to add a test, IMO.

partner_risk_insurance_security/views/res_partner_view.xml
toita86: I have tried to do 
```
<xpath expr="//page[@name='credit_insurance']" position="attributes">
                <attribute name="attrs">{'readonly': [('is_full_access_user', '=', False)]}</attribute>
            </xpath>
```
but it didn't work. I don't think that is possible to use the readonly attribute on a page

HekkiMelody: Fair enough, thanks for checking!

partner_risk_insurance_security/views/res_partner_view.xml
HekkiMelody: Sorry, I realize it was ambiguous. By "drop" I meant "remove" the line where you specify the priority entirely.

Anyway, non-blocking.

partner_risk_insurance_security/tests/test_partner_risk_insurance_security.py
```diff
@@ -0,0 +1,65 @@
...
+            {
+                "name": "Full Access User",
+                "login": "full_access_user",
+                "groups_id": [
+                    (
+                        6,
+                        0,
+                        [
+                            self.env.ref(
+                                "partner_risk_insurance_security.group_full_access_credit_policy_state"  # noqa: B950
```
HekkiMelody: **suggestion**(non-blocking): 

```suggestion
                                "partner_risk_insurance_security."
                                "group_full_access_credit_policy_state"
```

HekkiMelody (approved):
Code review, LGTM

carlosdauden (approved):
Code LGTM

Let's wait a bit before merging to see if the suggestions are heeded.

partner_risk_insurance_security/tests/test_partner_risk_insurance_security.py
toita86: Thank you for the suggestion!

carlosdauden:
Thanks, ready to merge

Issue #378, opened 2024-05-03
account_invoice_overdue_reminder : modifiing "invoicing contact" resets to zero the Total Residual in (currency) (BUG)


LESTRAT21:
##general summary of the issue
When you want to remind your customer, by mail, for his invoices that are overdue, you use the "Overdue Invoice Step" and you choose the "invoicing contact". If your contact has several related contact and you select one of them, the total residual is reset to zero. It should not be modified.

## Module
invoice_overdue_remider

## Describe the bug
look at the following video :
https://www.loom.com/share/f0d827c7d1944bfa9e172c38336205c7?sid=1d18bdb7-e95e-4494-839c-9279062dc33c

## To Reproduce
V14
The bug is not present in V16

1. Invoicing / Customers / Overdue Invoice Remind
2. Select your customer (it must be one with attached contacts) / Start
3. Change the invoicing contact (to select another e-mail for example)
4. In the e-mail below, the total residual is reset to zero.

**Expected behavior**
the total residual should not be reset to zero.

thomich:
seams lot of changes from 14 to 15 also:

[account_invoice_overdue_reminder.tar.gz](https://github.com/OCA/credit-control/files/15330323/account_invoice_overdue_reminder.tar.gz)

matthieusaison:
I propose this fix : 
https://github.com/OCA/credit-control/pull/406

PR #406, opened 2024-11-19
[14.0][FIX] account_invoice_overdue_reminder: wizard amount residual not computed

matthieusaison:
Resolve this issue https://github.com/OCA/credit-control/issues/378
(bug description in the issue)

### DIFF
--- a/account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
+++ b/account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
@@ -385,6 +385,20 @@ class OverdueReminderStep(models.TransientModel):
     def _reminder_type_selection(self):
         return self.env["overdue.reminder.action"]._reminder_type_selection()
 
+    # Fix bug:
+    # When removing a invoice in the interface the field "mail_body" is recomputed in
+    # "onchange mode". This mean that the field "invoice_ids" have New Record (NewId)
+    # instead of real record. The issue is that the field "amount_residual" do
+    # not work with New Record due to "if line.id" here:
+    # https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185
+    # /addons/account/models/account_move.py#L4029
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
weird bug. 
But LGTM.

LESTRAT21:
Thanks for having fixed this bug @matthieusaison

account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
```diff
@@ -385,6 +385,13 @@ class OverdueReminderStep(models.TransientModel):
     def _reminder_type_selection(self):
         return self.env["overdue.reminder.action"]._reminder_type_selection()
 
+    # fix bug:
+    # get_invoices() has to be use in mail template
+    # bug: In the wizard after delete an invoice to remind.
+    # The residuals values  are set to 0 in the generated mail.
```
sebastienbeau: Can you put more explanation ?
We spend time to understand it, it's important for other to understand the issue when migrating the module.

Maybe you can put something like

```
# When removing a invoice in the interface the field "mail_body" is recomputed in "onchange mode".
# This mean that the field "invoice_ids" have New Record (NewId) instead of real record.
# The issue is that the field "amount_residual" do not work with New Record
# due to "if line.id" here  https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/account/ models/account_move.py#L4029)

# And so the amount_residual will be always be recomputed by the onchange to 0 

# To avoid this issue we force in the template to use real invoices 
```

sebastienbeau (changes requested):
Please just update the comment

account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
matthieusaison: ok, great description, done

sebastienbeau (approved):
LGTM

PR #409, opened 2024-12-20
[16.0][FIX] account_credit_control: Cancel orders without permissions

dalonsofl:
### Module

account_credit_control

### Describe the bug

It is not possible to cancel quotations with no invoices created without Credit Control and Invoicing permissions

### To Reproduce

1. Create a sale order
2. Try to cancel

(image from runboat OCA)

image

image

image

**Affected versions:**

16.0

### DIFF
--- a/account_credit_control/models/account_move.py
+++ b/account_credit_control/models/account_move.py
@@ -35,14 +35,12 @@ class AccountMove(models.Model):
     def button_cancel(self):
         """Prevent to cancel invoice related to credit line"""
         # We will search if this invoice is linked with credit
-        cc_line_obj = self.env["credit.control.line"]
+        cc_line_obj = self.env["credit.control.line"].sudo()
         nondraft_domain = [
             ("invoice_id", "in", self.ids),
             ("state", "!=", "draft"),
         ]
-
         cc_nondraft_lines = cc_line_obj.search(nondraft_domain, limit=1)
-
         if cc_nondraft_lines:
             raise UserError(
                 _(


flotho (approved):
thanks for this, LGTM

StefanRijnhart (changes requested):
Fix looks good, but would you consider to add a small test to prevent this bug from reappearing after a refactoring?

enriquemartin (approved):

account_credit_control/models/account_move.py
```diff
@@ -40,20 +40,20 @@ def button_cancel(self):
             ("invoice_id", "in", self.ids),
             ("state", "!=", "draft"),
         ]
+        if self:
+            cc_nondraft_lines = cc_line_obj.search(nondraft_domain, limit=1)
```
len-foss: I wouldn't do this self check, but 
```
cc_line_obj = self.env["credit.control.line"].sudo()
```
and leave the rest identical.

len-foss (commented):
@StefanRijnhart how do you see a test for that? It's `sale` that calls this function with a user that doesn't have accounting rights. So this module should define an arbitrary object that will call cancel on an account move? I don't see how that's really doable.

Issue also affects other versions (at least the 17.0)

account_credit_control/models/account_move.py
dalonsofl: Done!

dalonsofl:
Hi @StefanRijnhart ! I agree with @len-foss , 

it is sale module which calls this function, i think it's not necessary to add test because of it. What do you think?

Thanks!

StefanRijnhart:
@len-foss @dalonsofl It's a public method for which there is a button in the view: https://github.com/odoo/odoo/blob/16.0/addons/account/views/account_move_views.xml#L609-L610.

To test the permission error, you can just modify the last lines of the existing `test_action_cancel` and call button_cancel with an unprivileged user. If you do, please improve those lines to use `self.assertReplacesRegex` with a snippet of the error message to make sure we are testing the right thing.

len-foss:
The line of code you point out explicitly has `groups="account.group_account_invoice`. Even it it's 'public', you can't expect a user that does not have this group to have this succeed. If it's the case, there's something very wrong about access rights.

My proposed fix is precisely that you let the cancellation happen or not _depending on the business logic intended by the module_. The reason why it should work for an unprivileged user (in most cases) is that the `account.move` object this is called on is empty, so there is nothing to do; the ACL is just misleading. If there  was indeed an object coming up, the unprivileged user should have the business exception.

StefanRijnhart:
@len-foss Good point! If the fact that button_cancel is called on an empty recordset, and we expect it to succeed for users that do not have any accounting group (because a quotation has no invoices), then we can emulate just that in a test: have an unprivileged user call `self.env["account.move"].button_cancel()` with a short comment explaining the reason for the test.

PR #417, opened 2025-01-09
[14.0][FIX][sale_financial_risk] compute of risk_sale_order write on commercial partner

Kev-Roche:
Compute of risk_sale_order write on commercial_partner_id, not on self. 
cc @sebastienbeau , @alexis-via

### DIFF
--- a/sale_financial_risk/models/res_partner.py
+++ b/sale_financial_risk/models/res_partner.py
@@ -26,7 +26,7 @@ class ResPartner(models.Model):
         risk_states = self.env["sale.order"]._get_risk_states()
         return self._get_risk_company_domain() + [
             ("state", "in", risk_states),
-            ("risk_partner_id", "in", self.mapped("commercial_partner_id").ids),
+            ("risk_partner_id", "in", self.ids),
         ]
 
     @api.depends(


sale_financial_risk/models/res_partner.py
```diff
@@ -26,7 +26,7 @@ def _get_risk_sale_order_domain(self):
         risk_states = self.env["sale.order"]._get_risk_states()
         return self._get_risk_company_domain() + [
             ("state", "in", risk_states),
-            ("risk_partner_id", "in", self.mapped("commercial_partner_id").ids),
+            ("risk_partner_id", "in", self.ids),
```
carlosdauden, line 29: Can you explain the problem this PR is trying to solve and the steps to reproduce it?
Thanks

carlosdauden (changes requested):

sale_financial_risk/models/res_partner.py
sebastienbeau: It's a tricky issue.


Let's explain

You open a contact "A" that have for commercial_partner_id the partner "B"
And you have an invoice of 1000€ on partner B.

You expect to have 1000€ of risk_sale_order on partner B and 0 on partner A.
If you try the module it have exactly this expected behavior.

But if you inspect the code you will see that odoo do a write when this field is computed (normally compute never call write method, and this raise issue with other module)


So let's going deeper !

When you open the form view of partner "A" the field "risk_sale_order" is computed.
So in the compute method in self you have "**record A**".

** Note: **
When you do a compute only record in "**self**" should have value assigned for two reason
- first is useless (odoo when to compute record A, it's not needed to compute record B)
- secondly it's broken as setting other value on a record in a compute will use a write instead of using the normal behaviour of computed field


**So what happen here:**

The domain is going to return the grouped amount of "sale.order.line" with the partner B (the commercial_partner_id)

Then the following code https://github.com/OCA/credit-control/blob/0b5c6eb96493187604a6af0fc4a2cf47398ccef1/sale_financial_risk/models/res_partner.py#L45-L56

Is going to set the "risk_sale_order" on partner B and 0 on partner A. But as we have in self only "record A" a write will be call on "partner B". Computing the value on partner B is useless as "partner A" will always have 0 as value and this break in our case security rule as some use do not have write access on "partner B" (and so the write raise an error)

PR #369, opened 2024-03-15
[17.0][MIG] account_credit_control: migrate to V17

astirpe:

PR #426, opened 2025-02-11
[17.0][FIX] account_credit_control: Render invoice details in emails correctly

carlos-lopez-tecnativa:
Before this commit, the existing code was obsolete because, in V17, the `body` field is computed, and `_onchange_template_id` no longer exists. This commit adapts the code to correctly include invoice details in the email body.

Complementary to PR https://github.com/OCA/credit-control/pull/369
TT54881
@Tecnativa @victoralmau @sergio-teruel @pedrobaeza could you please review this

### DIFF
--- a/account_credit_control/tests/test_credit_control_run.py
+++ b/account_credit_control/tests/test_credit_control_run.py
@@ -9,7 +9,7 @@ from dateutil import relativedelta
 
 from odoo import fields
 from odoo.exceptions import AccessError, UserError
-from odoo.tests import tagged
+from odoo.tests import RecordCapturer, tagged
 from odoo.tests.common import Form
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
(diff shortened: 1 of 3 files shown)


account_credit_control/wizard/mail_compose_message.py
```diff
@@ -3,22 +3,26 @@
...
     _inherit = "mail.compose.message"
 
-    def _onchange_template_id(self, template_id, composition_mode, model, res_id):
-        res = super()._onchange_template_id(
-            template_id=template_id,
-            composition_mode=composition_mode,
-            model=model,
-            res_id=res_id,
-        )
+    @api.depends("composition_mode", "model", "res_domain", "res_ids", "template_id")
```
pedrobaeza: Why adding again all the depends?

pedrobaeza (commented):
Can you add a regression test?

account_credit_control/wizard/mail_compose_message.py
carlos-lopez-tecnativa: Fixed

sergio-teruel (approved):

account_credit_control/tests/test_credit_control_run.py
```diff
@@ -216,6 +216,55 @@ def test_wiz_credit_control_emailer(self):
...
+        marker = self.env["credit.control.marker"].create(
+            {"name": "to_be_sent", "line_ids": [(6, 0, control_lines.ids)]}
+        )
+        marker.mark_lines()
+        emailer_obj = self.env["credit.control.emailer"]
+        wiz_emailer = emailer_obj.create({})
+        wiz_emailer.line_ids = control_lines
+        with RecordCapturer(self.env["credit.control.communication"], []) as capture:
+            wiz_emailer.email_lines()
+        new_communication = capture.records
```
pedrobaeza, line 244: Taking `capture` outside the with may lead to errors, isn't it?

carlos-lopez-tecnativa: No, in many examples from Odoo and OCA, this is a common practice. For example:

- https://github.com/odoo/odoo/blob/b1b9266ef25c5a25f752f83d84a87e71192763a2/addons/website_event_crm/tests/test_event_registration.py#L110-L138
- https://github.com/odoo/odoo/blob/b1b9266ef25c5a25f752f83d84a87e71192763a2/addons/website_crm/tests/test_website_crm.py#L44-L46
- https://github.com/OCA/field-service/blob/47c9a51eb8beadf2b2b0591dc8ae3d92e0f4a38b/fieldservice_stage_server_action/tests/test_fsm_order_run_action.py#L36-L38
- https://github.com/search?q=org%3AOCA%20RecordCapturer&type=code

pedrobaeza (approved):
/ocabot merge patch

PR #428, opened 2025-02-13
[16.0][IMP] sale_payment_sheet_financial_risk: Discard payment sheets with paid invoices

carlosdauden:
@Tecnativa TT55087

ping @sergio-teruel @CarlosRoca13

### DIFF
--- a/sale_payment_sheet_financial_risk/models/res_partner.py
+++ b/sale_payment_sheet_financial_risk/models/res_partner.py
@@ -41,6 +41,7 @@ class ResPartner(models.Model):
         )
         return [
             ("partner_id", "in", commercial_partners.ids),
+            ("invoice_id.payment_state", "!=", "paid"),
             "|",
             ("state", "=", "open"),
             ("statement_line_id.state", "=", "open"),


sergio-teruel (approved):
Great !! Thanks!!

CarlosRoca13 (approved):

sale_payment_sheet_financial_risk/models/res_partner.py
```diff
@@ -41,6 +41,7 @@ def _get_risk_sale_payment_sheet_domain(self):
         )
         return [
             ("partner_id", "in", commercial_partners.ids),
+            ("invoice_id.payment_state", "!=", "paid"),
```
pedrobaeza, line 44:
```suggestion
            ("invoice_id.payment_state", "not in", ("in_payment", "paid")),
```

pedrobaeza (changes requested):

sale_payment_sheet_financial_risk/models/res_partner.py
carlosdauden: The suggested change is not exactly what we are looking for, if it is ‘in_payment’ it should be included because the balance is handled differently, as it is with open invoices for example.
This change is only to cover a rare case where the payment_sheet is left open, but the invoice has already been paid.

pedrobaeza (approved):
Ok then

/ocabot merge patch

PR #429, opened 2025-02-13
[17.0][FIX] account_credit_control: Avoid error when opening or creating a mail template

carlos-lopez-tecnativa:
When a user with the `group_account_credit_control_manager` group tries to select a `mail.template`, an access error is raised because `model_id(ir.model)` has permissions restricted to `Administration/Access Rights`. Instead of using a many2one field, the `model` field is used, as it is `related` and `stored`.

image

TT55104
@Tecnativa @pedrobaeza @sergio-teruel could you please review this?

### DIFF
--- a/account_credit_control/views/credit_control_policy.xml
+++ b/account_credit_control/views/credit_control_policy.xml
@@ -45,6 +45,7 @@
                                         <group colspan="2">
                                             <field
                                                 name="email_template_id"
+                                                context="{'default_model': 'credit.control.communication'}"
                                                 colspan="2"
                                             />
                                             <field name="custom_text" colspan="2" />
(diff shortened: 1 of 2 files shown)


account_credit_control/views/credit_control_policy.xml
```diff
@@ -45,6 +45,7 @@
                                         <group colspan="2">
                                             <field
                                                 name="email_template_id"
+                                                options="{'no_create': True}"
```
pedrobaeza: Why this? Someone with permissions may want to create the template from here.

carlos-lopez-tecnativa: A user may have permission to create `mail.template`, but when trying to select the `model(Applies to)`, the error appears again.

image

carlos-lopez-tecnativa: I can set `default_model: "credit.control.communication"` by context, but if the user changes or opens the field, the error appears again. Let me know if I should still add the default.
image

pedrobaeza: Well, I think you have to pre-fill it via context default.

carlos-lopez-tecnativa: Please check again and let me know if it's OK for you.

pedrobaeza (approved):
Yes, I think it's OK this way, and we can fast-track for avoiding the access error:

/ocabot merge patch

PR #434, opened 2025-02-25
[16.0][FIX] account_payment_return_financial_risk + sale_financial_risk + sale_financial_risk_info: Post-install test + fallback to load CoA 

victoralmau:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with the error:

`odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale`

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/account_payment_return_financial_risk/tests/test_payment_return_risk.py
+++ b/account_payment_return_financial_risk/tests/test_payment_return_risk.py
@@ -2,16 +2,26 @@
 # Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo.tests import Form, common
+from odoo.tests import Form, common, tagged
 
 from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
 
 
+@tagged("post_install", "-at_install")
 class TestPartnerPaymentReturnRisk(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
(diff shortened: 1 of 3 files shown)


victoralmau:
What is the reason for the tests not finishing (although there is no error)? Any suggestions?

pedrobaeza:
I don't know... they seem frozen. I have restarted them again.

pedrobaeza:
I have canceled it again, as it always stuck. It's very weird. You can try to make non installable all modules and enable one by one until we get which one gives the problem.

victoralmau:
I have finally been able to "`replicate`" the crash locally (without any message indicating exactly what the problem is). 

I have confirmed that this does not happen at https://github.com/OCA/credit-control/pull/436.

I have tried making this change https://github.com/OCA/sale-workflow/pull/3595 and including it in this PR to confirm that the CI is now working correctly but I honestly don't understand exactly why.

victoralmau:
Finally this is ready to merge.

pedrobaeza (approved):
The same as done across 16.0, so fast-tracking it having green CI:

/ocabot merge patch

PR #418, opened 2025-01-10
[18.0][MIG] account_credit_control: Migration to 18.0

kevinkhao:

PR #437, opened 2025-03-10
[18.0][MIG] account_credit_control: Migration to version 18.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/credit-control/pull/418
TT54066
@Tecnativa @pedrobaeza @sergio-teruel @victoralmau @StefanRijnhart  could you please review this?


victoralmau:
Please, cherry-pick https://github.com/OCA/credit-control/pull/438 to commit history (before migration commit).

StefanRijnhart:
@victoralmau Just curious, but does the order of the commits matter?

victoralmau:
> @victoralmau Just curious, but does the order of the commits matter?

IMO there are several reasons:
- Readability: adding the commit in the right place will help to know quickly in the commit history whether that change was made before the migration or after.
- Pre-commit: The right thing to do is to add the commit before the pre-commit commit so that if it is necessary to apply any change it is “separated” in that commit and does not confuse.
- Migration: Adding the commit before will help to apply the changes that are necessary for the migration in the migration commit and thus to know the changes that really have been done in the migration (only those).

carlos-lopez-tecnativa:
> Please, cherry-pick #438 to commit history (before migration commit).

Done

account_credit_control/security/account_security.xml
```diff
@@ -0,0 +1,93 @@
...
+        </record>
+    </data>
+    <data noupdate="1">
+        <record id="credit_control_run_rule" model="ir.rule">
+            <field name="name">Credit Control Run</field>
+            <field name="model_id" ref="model_credit_control_run" />
+            <field name="global" eval="True" />
+            <field
+                name="domain_force"
+            >['|',('company_id','=',False),('company_id','in',company_ids)]</field>
```
StefanRijnhart: This should be better for performance (here and further below).

```suggestion
            >['|', ('company_id','in',company_ids + [False])]</field>
```

account_credit_control/models/credit_control_policy.py
```diff
@@ -270,7 +271,7 @@ def _generate_policy_lines_with_max_level(
         return policy_lines_generated
 
     @api.model
-    def _name_search(self, name, domain=None, operator="ilike", limit=None, order=None):
+    def _search_display_name(self, operator, value):
```
StefanRijnhart, line 274: Maybe a test, given that the change went unnoticed on the original PR because there was no test before?

carlos-lopez-tecnativa: Done

account_credit_control/security/account_security.xml
carlos-lopez-tecnativa: Done

StefanRijnhart (approved):
Great, thanks!

account_credit_control/security/account_security.xml
```diff
@@ -41,47 +41,47 @@
             <field name="global" eval="True" />
             <field
                 name="domain_force"
-            >['|',('company_id','=',False),('company_id','in',company_ids)]</field>
+            >[('company_id','in',company_ids + [False])]</field>
```
pedrobaeza, line 44: This requires migration script

account_credit_control/models/mail_message.py
```diff
@@ -1,25 +0,0 @@
-# Copyright 2020 Creu Blanca
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from odoo import models
-
-
-class MailMessage(models.Model):
-    _inherit = "mail.message"
-
-    def message_format(self, format_reply=True, msg_vals=None):
```
pedrobaeza: Why removing this? It's a pity that there's no docstring stating its purpose though.

account_credit_control/security/account_security.xml
carlos-lopez-tecnativa: Done, I updated the domain for `credit_control_communication_rule` because the field is required. So, as mentioned in another PR, `[False]` is unnecessary.

account_credit_control/models/mail_message.py
carlos-lopez-tecnativa: Because this function no longer exists, I asked the same question but couldn't find a similar function. Also, I don't understand what its code does. If you know, please let me know.

StefanRijnhart: Read downwards from here: https://github.com/OCA/credit-control/pull/86#issuecomment-689408353

Currently, it may prevent the header from being rendered in the email layout: https://github.com/odoo/odoo/blob/18.0/addons/mail/data/mail_templates_email_layouts.xml#L25. If that's the case, that might not be what we want.

The field `is_discussion` is still present on mail.message, and it is equivalent to subtype == mt_comment in various places (e.g. https://github.com/odoo/odoo/blob/18.0/addons/mail/models/mail_thread.py#L3570) so if we think the gray layout in the Odoo frontend is still an issue we *could* rework the feature but I'm not sure.

carlos-lopez-tecnativa: @StefanRijnhart, thanks for the links. After investigating further, I noticed that this is no longer necessary. The reason is that during the [migration to V17](https://github.com/OCA/credit-control/pull/369/commits/0c8e247ddcdacce97def6da9ebfb7848fdf06823#diff-ab2ca28045e4ca81ddb0cacf900797263a04da3c88c8f3f46b93ebeca3b2d9bbR222), the function `message_post_with_template` was replaced by `message_mail_with_source`, which now handles the message as an email by passing `'composition_mode': 'mass_mail'`.
https://github.com/odoo/odoo/blob/4af02a87266883f3551886b8def2de261cb69cd3/addons/mail/models/mail_thread.py#L2506

So, when the `mail.compose.message` has `composition_mode == "mass_mail"`, the `subtype_id` is set to `False` but is considered as an email.

https://github.com/odoo/odoo/blob/4af02a87266883f3551886b8def2de261cb69cd3/addons/mail/wizard/mail_compose_message.py#L938-L941

You can confirm this by comparing the `Type` and `Subtype` of messages in V18 and earlier versions.

V18
image

V12
image

carlos-lopez-tecnativa: In the chatter, the message is now displayed as a mail.
image

I updated the code because during the migration done by another user, the chatter section was not replaced correctly. It only had `<div class="oe_chatter" />`, but it should have been `<chatter />`. As a result, the chatter was not rendered correctly. Now everything is fine

carlosdauden (approved):

account_credit_control/models/mail_message.py
StefanRijnhart: Nice, thanks for the update!

carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #438, opened 2025-03-20
[17.0][IMP+FIX] account_credit_control: Defining the right data in the communication email

victoralmau:
<ins>Defining the right data in the communication email.</ins>

Changes done:
- Defined date_entry for the Invoice date column.
- Define amount_due for Invoice amount column
- Set balance_due for the Amount Due column

This data is the same as already shown in the Credit Control Summary report.

<ins>Add `date_entry` field to lines views.</ins>

<ins>Define the appropriate value in Invoice number for moves.</ins>

<ins>Change class=text-right to class=text-end in the report for correct visualization.<ins>

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT55608

### DIFF
--- a/account_credit_control/models/credit_control_communication.py
+++ b/account_credit_control/models/credit_control_communication.py
@@ -194,7 +194,10 @@ class CreditControlCommunication(models.Model):
         table_content = "<br/><h3>%s</h3>" % _("Invoices summary")
         table_content += "<table style='%s'><tr>%s</tr>" % (table_style, tr_content)
         for line in self.credit_control_line_ids:
-            tr_content = "<td style='%s'>%s</td>" % (th_style, line.invoice_id.name)
+            name = line.invoice_id.name
+            if line.move_line_id.ref and line.move_line_id.ref != name:
+                name += f" ({line.move_line_id.ref})"
+            tr_content = "<td style='%s'>%s</td>" % (th_style, name)
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
                 line.invoice_id.payment_reference or "",
--- a/account_credit_control/models/credit_control_communication.py
+++ b/account_credit_control/models/credit_control_communication.py
@@ -201,7 +201,7 @@ class CreditControlCommunication(models.Model):
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
-                format_date(self.env, line.invoice_id.invoice_date),
+                format_date(self.env, line.date_entry),
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
@@ -209,11 +209,11 @@ class CreditControlCommunication(models.Model):
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
-                format_amount(self.env, line.invoice_id.amount_total, line.currency_id),
+                format_amount(self.env, line.amount_due, line.currency_id),
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
-                format_amount(self.env, line.amount_due, line.currency_id),
+                format_amount(self.env, line.balance_due, line.currency_id),
             )
             table_content += "<tr>%s</tr>" % tr_content
         table_content += "</table>"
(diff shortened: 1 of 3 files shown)


account_credit_control/models/credit_control_communication.py
```diff
@@ -194,26 +194,30 @@ def _get_credit_control_communication_table(self):
         table_content = "<br/><h3>%s</h3>" % _("Invoices summary")
         table_content += "<table style='%s'><tr>%s</tr>" % (table_style, tr_content)
         for line in self.credit_control_line_ids:
-            tr_content = "<td style='%s'>%s</td>" % (th_style, line.invoice_id.name)
+            if line.invoice_id.move_type == "entry":
+                name = line.move_line_id.ref
```
pedrobaeza: Combine both move name and ref. We can't stick to a particular case.

account_credit_control/models/credit_control_communication.py
```diff
@@ -194,26 +194,30 @@ def _get_credit_control_communication_table(self):
...
+                format_date(self.env, line.date_entry),
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
                 format_date(self.env, line.date_due),
             )
             tr_content += "<td style='%s'>%s</td>" % (
                 th_style,
-                format_amount(self.env, line.invoice_id.amount_total, line.currency_id),
+                format_amount(self.env, line.amount_due, line.currency_id),
```
pedrobaeza, line 215: This is not correct. It should be the total of the line. Please check how it's done in the report.

account_credit_control/views/credit_control_line.xml
```diff
@@ -23,6 +23,7 @@
                         </group>
                         <group>
                             <field name="date" readonly="state != 'draft'" />
+                            <field name="date_entry" />
```
pedrobaeza, line 26: Why?

victoralmau: This is a useful information, remember that there are 3 dates:
- Date (`date`): Date of the credit control line.
- Entry date (`date_entry`): The date of the journal item (`move_line_id.date`).
- Due date (`date_due`): Due date.

It is important to have in context all the dates because each one represents different things and will probably be different. Maybe a hidden optional field?

account_credit_control/models/credit_control_communication.py
victoralmau: This change is precisely to indicate the same data shown in the report. https://github.com/OCA/credit-control/blob/df877b848a5653b2b83acb7bb80b4c2e78f15f67/account_credit_control/report/report_credit_control_summary.xml#L58

pedrobaeza: Uhm, it sounds weird. Both are the residual. Where is the total expressed?

account_credit_control/views/credit_control_line.xml
pedrobaeza: OK, understood.

account_credit_control/models/credit_control_communication.py
victoralmau: Only the `amount_due` and `balance_due` fields exist, `amount_due` seems to represent the appropriate value. https://github.com/OCA/credit-control/blob/df877b848a5653b2b83acb7bb80b4c2e78f15f67/account_credit_control/models/credit_control_line.py#L181

pedrobaeza: OK, the name is totally misleading.

pedrobaeza (approved):

carlos-lopez-tecnativa (changes requested):
LGTM. Additionally, please update the code to Bootstrap 5 by changing `text-right` to `text-end` to correctly align the amounts.
image

https://github.com/OCA/credit-control/blob/3e134266be86bb37b6fe3977738b7a16b3bd489c/account_credit_control/report/report_credit_control_summary.xml#L61

carlos-lopez-tecnativa (approved):

PR #440, opened 2025-03-27
[18.0][FIX] sale_financial_risk: avoid compute for downpayment lines

### LINKED ISSUE #439: 18.0 sale_financial_risk: Stacktrace when creating down payment invoice
Error when creating a downpayment invoice from a Sale Order

## Module
sale_financial_risk

## Describe the bug
Impossible to create a downpayment invoice from a Sale Order.

## To Reproduce
**Affected versions**: 18.0

1. Go to a Sale Order that is not invoiced (example S00026)
2. Select 'Create Invoice'
3. Choose "Down Payment" (percentage or fixed)
4. Select "Create Draft"
5. Produces stacktrace

**Expected behavior**
Invoice draft should be created

**Additional context**
Downpayment lines do not have a UOM assigned, which leads to an error here:
https://github.com/OCA/credit-control/blob/1ae113f85a6d6a66e66d6c592fae4cb1d63ccfe0/sale_financial_risk/models/sale.py#L104

Solution:
The `_compute_risk_amount` function should probably ignore downpayment lines:
https://github.com/OCA/credit-control/blob/1ae113f85a6d6a66e66d6c592fae4cb1d63ccfe0/sale_financial_risk/models/sale.py#L97

Stacktrace:
```
File "/odoo/addons/sale_financial_risk/models/sale.py", line 106, in _compute_risk_amount
    risk_qty = float_round(
               ^^^^^^^^^^^^
  File "/odoo/tools/float_utils.py", line 70, in float_round
    rounding_factor = _float_check_precision(precision_digits=precision_digits,
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/tools/float_utils.py", line 35, in _float_check_precision
    assert precision_rounding > 0,\
           ^^^^^^^^^^^^^^^^^^^^^^
AssertionError: precision_rounding must be positive, got 0.0
```

tobiaszehntner:
Fixes https://github.com/OCA/credit-control/issues/439

### DIFF
--- a/sale_financial_risk/models/sale.py
+++ b/sale_financial_risk/models/sale.py
@@ -97,7 +97,11 @@ class SaleOrderLine(models.Model):
     def _compute_risk_amount(self):
         risk_states = self.env["sale.order"]._get_risk_states()
         for line in self:
-            if line.state not in risk_states or line.display_type:
+            if (
+                line.state not in risk_states
+                or line.display_type
+                or line.is_downpayment
+            ):
                 line.risk_amount = 0.0
                 continue
             qty = line.product_uom_qty
--- a/sale_financial_risk/tests/test_partner_sale_risk.py
+++ b/sale_financial_risk/tests/test_partner_sale_risk.py
@@ -134,16 +134,28 @@ class TestPartnerSaleRisk(BaseCommon):
         self.assertTrue(self.partner.risk_exception)
         # If we create and validate an invoice from the sale order then the
         # amount to be invoiced must be 0 and risk_exception must be False
-        inv_wiz = (
+        # Using down payment plus normal invoice to test both options
+        context = {"active_ids": [self.sale_order.id]}
+        downpayment = (
             self.env["sale.advance.payment.inv"]
-            .with_context(**{"active_ids": [self.sale_order.id]})
-            .create({})
+            .with_context(**context)
+            .create(
+                {
+                    "advance_payment_method": "fixed",
+                    "fixed_amount": 50,
+                }
+            )
         )
-        inv_wiz.create_invoices()
+        downpayment.create_invoices()
+        payment = (
+            self.env["sale.advance.payment.inv"].with_context(**context).create({})
+        )
+        payment.create_invoices()
+
         self.assertAlmostEqual(self.partner.risk_invoice_draft, 115.0)
         self.assertAlmostEqual(self.partner.risk_sale_order, 0)
-        invoice = self.sale_order.invoice_ids
-        invoice.with_context(bypass_risk=True).action_post()
+        invoices = self.sale_order.invoice_ids
+        invoices.with_context(bypass_risk=True).action_post()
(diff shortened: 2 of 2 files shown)


acsonefho (changes requested):
Please can you add a test about your case :pray:

acsonefho (approved):

rafaelbn:
@moduon MT-10250 please @Shide could you kindly review? Thank you ! 😄 ❤️

Shide (approved):
Code lgtm

pedrobaeza:
Shouldn't down payment lines reduce the risk?

tobiaszehntner:
> Shouldn't down payment lines reduce the risk?

I guess it would reduce the risk on the total of the SO, however, this function computes the risk on the line. And I would assume the risk of a downpayment line is 0?

However I lack the business knowledge for this, so I'm happy if others could weigh in on that and I can fix anything being proposed.

tobiaszehntner:
> @rafaelbn The rebase process failed, because command `git push --force acsone tmp-pr-440:18.0-fix-sale_financial_risk` failed with output:
> 
> ```
> remote: Permission to acsone/credit-control.git denied to OCA-git-bot.
> fatal: unable to access 'https://github.com/acsone/credit-control/': The requested URL returned error: 403
> ```

I rebased manually

pedrobaeza:
The risk for the line should be negative for the amount of the down payment, so the sum of this line across the order will reduce the total risk.

tobiaszehntner:
> The risk for the line should be negative for the amount of the down payment, so the sum of this line across the order will reduce the total risk.

Quickly checked and think we'd need specific test cases. I cannot dedicate time for that at the moment since our customer doesn't use this but was only blocked by the stacktrace. So if there are others up for improving this, please do 🙏

Otherwise maybe the suggestion would be more of a separate improvement, while this fix could unblock users?

carlosdauden:
I don't quite understand this thread...

Theoretically the origin of the problem was this:
https://github.com/OCA/credit-control/issues/439
because it was missing a commit (which had already been incorporated some time ago in previous versions)
@AaronHForgeFlow solution to the problem:
https://github.com/OCA/credit-control/pull/450

I think this PR should be closed and if you consider it appropriate to open a new one with the supposed new problem and the solution to correct it...

carlosdauden (changes requested):
I don't quite understand this thread...

Theoretically the origin of the problem was this:
https://github.com/OCA/credit-control/issues/439
because it was missing a commit (which had already been incorporated some time ago in previous versions)
@AaronHForgeFlow solution to the problem:
https://github.com/OCA/credit-control/pull/450

I think this PR should be closed and if you consider it appropriate to open a new one with the supposed new problem and the solution to correct it...

tobiaszehntner:
@carlosdauden perfect, thank you for pointing this out. Fixed with https://github.com/OCA/credit-control/pull/450. Closing

PR #441, opened 2025-03-27
[17.0][FIX] account_financial_risk: Display the Financial Risk page correctly...

celm1990:
...when the contact is a person but does not have a parent_id. 
During migration to V17, the conversion from attribute to Python expression was incorrect. As a result, this page became visible when a contact is a person but has a parent_id. The correct logic should display the field only when the contact is a person and does not have a parent_id.

Before this commit
image

@pedrobaeza @carlosdauden @sergio-teruel could you please review this?

After this commit
image

### DIFF
--- a/account_financial_risk/views/res_partner_view.xml
+++ b/account_financial_risk/views/res_partner_view.xml
@@ -12,7 +12,7 @@
                 <page
                     name="financial_risk"
                     string="Financial Risk"
-                    invisible="not is_company and not parent_id"
+                    invisible="not is_company and parent_id"
                     groups="account_financial_risk.group_account_financial_risk_user"
                 >
                     <group name="risk_general">


carlosdauden (approved):

account_financial_risk/views/res_partner_view.xml
```diff
@@ -12,7 +12,7 @@
                 <page
                     name="financial_risk"
                     string="Financial Risk"
-                    invisible="not is_company and not parent_id"
+                    invisible="not is_company and parent_id"
```
pedrobaeza, line 15: Why not simply?

```suggestion
                    invisible="parent_id"
```

carlosdauden: No, because if it is_company the commercial_partner_id is the same contact even if it has a parent_id set, so it must be visible in that case.

pedrobaeza (approved):
/ocabot merge patch

PR #448, opened 2025-04-22
[18.0][FIX] account_financial_risk: better format res_partner_view

gdgellatly:
In testing this module for v18 a few view inconsistencies became apparent.

- Extra border from oe_subtotal_footer on top or below.
- Groups a bit different in the way the two were declared.
- Buttons not inline with labels.
- Credit currency field was declared inside a div with label, much better outside.

This is a quick fix to make it presentable. In future we probably need to stop using oe_subtotal_footer and look to custom css.

Before:
image

image

After:
image

image

### DIFF
--- a/account_financial_risk/views/res_partner_view.xml
+++ b/account_financial_risk/views/res_partner_view.xml
@@ -21,7 +21,7 @@
                                 col="3"
                                 class="oe_subtotal_footer"
                                 string="General Limits"
-                                style="float: left !important;"
+                                style="float: left !important; border-top: none;"
                             >
                                 <field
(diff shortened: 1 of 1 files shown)


gdgellatly:
Note the button for sales orders is in PR #449 so you can see in screenshot the difference. One screenshot from runboat, the other from enterprise but tbh they look pretty much the same.

EDIT: Screenshot from runboat

image

rafaelbn (approved):
Thank you @gdgellatly !

👍🏼 

@moduon MT-MT-4970 ping @Shide

carlosdauden (approved):

PR #449, opened 2025-04-22
[18.0][FIX] sale_financial_risk: better format res_partner_view

gdgellatly:
Follows on from #448 to raise the sale order limit button to be inline.


rafaelbn (approved):
Thank you @gdgellatly !

👍🏼

@moduon MT-4970 ping @Shide

carlosdauden (approved):

PR #451, opened 2025-05-13
[18.0][MIG] account_invoice_overdue_reminder: Migration to 18.0

ByteMeAsap:


alexis-via:
Your migration commit to v18 has removed the PR that added refunds in the reminder email : https://github.com/OCA/credit-control/pull/346
The initial version of the module considered that refunds should be reconciled with invoices and the reminder email should only contain invoices (and there was a blocking error if odoo found a customer refunds that was not fully reconciled).
But some contributors wanted to have the unreconciled refunds listed with negative amount in the reminder email, which would impact the total to pay. Their argument was that it's not always possible to reconcile a refund with a specific invoice (for exemple for end-of-year discounts granted as a customer refund). Their contribution was merged and, unless decided otherwise, it should be kept.
So I'll propose another mig PR that include the support for unreconciled refunds.

ByteMeAsap:
@alexis-via  Thanks for pointing this put, I will propose a new PR

PR #455, opened 2025-05-13
[17.0][IMP] account_financial_risk: Avoid same priority than account.view_move_line_pivot

carlosdauden:
because account.action_account_moves_all have not defined view and it may be the case that the one in this module is used when it should not be.

FW-Port https://github.com/OCA/credit-control/pull/422

@Tecnativa TT54575


carlosdauden:
ping @sergio-teruel @victoralmau

sergio-teruel (approved):

victoralmau (approved):

PR #457, opened 2025-06-04
[16.0][FIX] account_credit_control: show balance_due instead of amount_due

TelmoSuarez:


TelmoSuarez:
Fixed in https://github.com/OCA/credit-control/pull/501

PR #464, opened 2025-07-02
[18.0][MIG] account_credit_control_attach_invoice: Migration to 18.0

victorvermot:


account_credit_control_attach_invoice/models/report.py
```diff
@@ -12,8 +12,15 @@ class IrActionsReport(models.Model):
     _inherit = "ir.actions.report"
 
     def _render_qweb_pdf(self, report_ref, res_ids=None, data=None):
-        reports = ["account_credit_control.report_credit_control_summary"]
-        if report_ref in reports and self.env.user.company_id.report_to_attach_id:
+        if isinstance(report_ref, models.BaseModel):
+            report_xml_id = report_ref.xml_id
+        else:
+            report_xml_id = report_ref
```
ivantodorovich: The safest way to deal with the fact that `report_ref` can be many things, is to call `_get_report`:
https://github.com/odoo/odoo/blob/d33d10e6297d6e8dc9521cdae86e5bd3e8d7d55b/odoo/addons/base/models/ir_actions_report.py/#L647-L655

So, something like this:

```suggestion
        report = self._get_report(report_ref)
        report_xml_id = report.xml_id
```

ivantodorovich (approved):

account_credit_control_attach_invoice/models/report.py
victorvermot: You are the best @ivantodorovich thanks for the tip !!

cyrilmanuel (approved):

leemannd (approved):

account_credit_control_attach_invoice/models/report.py
```diff
@@ -0,0 +1,69 @@
...
+    def _render_qweb_pdf(self, report_ref, res_ids=None, data=None):
+        report = self._get_report(report_ref)
+        report_xml_id = report.xml_id
+        reports = [
+            "account_credit_control.report_credit_control_summary",
+            "account_credit_control.credit_control_summary",
+        ]
+        if report_xml_id in reports and self.env.user.company_id.report_to_attach_id:
+            io_list = []
+            for cred_ctr in self.env["credit.control.communication"].browse(res_ids):
```
yvaucher: Only nickping, I do not consider this as important for my review:

`cred_ctr` is no intuitive as the object is a communication.

What about `comm` as in other `account_credit_control` modules?

(this would make the next variable `comm_pdf`)

account_credit_control_attach_invoice/models/res_company.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    report_to_attach_id = fields.Many2one(
```
yvaucher: This field name and string are too generic. A user might never find it.


What kind of report and what it will be attached to is important.
Would it also be possible to filter to report for invoices?

Shouldn't it be on the policy instead?

yvaucher (changes requested):
I'm mostly concerned about the parameter to set which report to attach as it lacks guidance for the user to find it and set it.

A better naming and a clear how to with the exact name of the field string would help is `USAGE.rst`.

Kimkhoi3010:
Hi @victorvermot, I created [PR](https://github.com/camptocamp/credit-control/pull/1) to improve documentation a bit (no change on code itself)

yvaucher:
@victorvermot Can you add @Kimkhoi3010 changes?

victorvermot:
@yvaucher Done. I also integrated your feedback and changed the string of the field ! Thanks a lot for the pr @Kimkhoi3010

account_credit_control_attach_invoice/models/res_company.py
yvaucher: Can you rename this field to something like `credit_control_report_to_attach_id` ?

yvaucher (approved):
@victorvermot Thanks for the changes

PR #430, opened 2025-02-18
[17.0][IMP] account_credit_control: Send communication emails in background

sergio-teruel:
cc @Tecnativa  TT54594

ping @carlosdauden @CarlosRoca13

PR #467, opened 2025-07-09
[18.0][FIX] account_credit_control: correctly update control lines status

len-foss:
This PR  is a forward port of https://github.com/OCA/credit-control/pull/447, and addresses the comment https://github.com/OCA/credit-control/pull/430/files#r1963626501

The state of the control lines should be changed after sending the email, but this wasn't done because there was a check on the subtype. 

Another bug that affected this lack of state change was: https://github.com/OCA/credit-control/pull/426/
Because related values weren't flushed, the lines weren't linked to the communication, and so the postprocess would never update their state.


pedrobaeza:
It seems you should squash the first 2 commits together, and why do you need the so called "flexibility" in the tests? You are not explaining the reasons in the commit.

len-foss:
@pedrobaeza honestly I'm not very sure. 
If you look at  https://github.com/OCA/credit-control/blob/c573021af070b1572c1187c053f38fab32322a54/account_credit_control/models/credit_control_policy.py#L229-L230, there is no root tag for the test. The regex assumed a `<p>`, and I got a `<span>`. I don't think this was relevant to the test (I could add this to the commit message indeed).
This problem wasn't triggered on the runboat, only on my local run (hence why I wrote it's more flexible).

I would not squash the 2 first commits because the second one should be cherry-picked for 17.0. I've done it in https://github.com/OCA/credit-control/pull/468

pedrobaeza:
Yes, please include in the commit message.

len-foss:
@pedrobaeza 
There's a way worse issue with the https://github.com/OCA/credit-control/pull/430, that is explained in details below. 


Commit:
https://github.com/odoo/odoo/commit/543a3ad204232cc67ae9e099f43960b664e8de19

Contains the comment that 
```
# If the value can't be parsed to an integer then it's computed in an automated way to
# half the size of db_maxconn because while most requests won't borrow cursors concurrently
# there are some exceptions where some controllers might allocate two or more cursors.
```

So a thread is generally assumed to take at most 2 cursors. 
Because of this, it is essentially a very bad idea to open an unlimited number of cursors.
(On a decently sized DB, about 400 lines would be processed in parallel...)

Because things are launched in their own cursor, this doesn't impact the main request, 
so the run is marked as done and all lines are marked as queued, 
despite the emails never being scheduled.

There is already a way to manage this kind of cases in the OCA that is called `queue_job`. 
Therefore this is the module that should handle this. 

If for some reason one would still want the threaded code nonetheless, 
it should be put in a custom project module, as it relies on some environment hypothesis 
that can't be tested from within the code. 

Note that one test is split into two; although it is better to avoid code duplication, 
the test was cheating the setup, and therefore not representative of the real user flow is 
when users click on the interface buttons.

pedrobaeza:
@sergio-teruel can you check @len-foss' last comment?

account_credit_control/tests/test_credit_control_run.py
```diff
@@ -288,30 +290,42 @@ def test_sent_email_invoice_detail(self):
...
-        emailer_obj = self.env["credit.control.emailer"].with_context(
-            domain_notifications_email="test@example.com"
-        )
-        wiz_emailer = emailer_obj.create({})
+        wiz_emailer = self.env["credit.control.emailer"].create({})
         wiz_emailer.line_ids = control_lines
+        self.env.user.company_id.email = "test@example.com"
         with RecordCapturer(self.env["credit.control.communication"], []) as capture:
-            wiz_emailer.email_lines()
+            wiz_emailer.with_context(queue_job__no_delay=True).email_lines()
```
yvaucher, line 297: Is it a leftover before you created a separate module for queue_job? or is it intended to make the module compatible in unit tests?

If it's for compatibility, can you add a small comment it'l for account_credit_control_queue_job ?

account_credit_control_queue_job/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2025 360ERP (<https://www.360erp.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Account Credit Control",
```
yvaucher, line 5: Did you had a look at :
https://github.com/OCA/credit-control/tree/12.0/account_credit_control_queue_job/

It was adding some other actions in queue job.

For instance emails were processed by batch.
It also adds a new state `sending` on the credit lines.

account_credit_control/models/credit_control_communication.py
```diff
@@ -233,36 +232,20 @@ def _generate_emails(self):
...
-        else:
-            self._send_communications_by_email()
-
-    def send_mails_threaded(self, record_ids):
-        with registry(self._cr.dbname).cursor() as cr:
-            self = self.with_env(self.env(cr=cr))
-            communications = self.browse(record_ids)
-            communications._send_communications_by_email()
+        # in account_credit_control_queue_job, override this method
+        # to loop over self and call _send_communications_by_email with delay
```
yvaucher, line 236: Indeed, looks more robust to me to use the queue job.

account_credit_control/tests/test_credit_control_run.py
len-foss: It's in the commit message. Indeed if you run the tests for account_credit_control you don't need it, but since on the standard workflow is to install modules then run the tests, my implementation breaks this flow.

account_credit_control_queue_job/__manifest__.py
len-foss: Thank you for letting me know.
It's generally a bad idea to add a new state to a selection field, and I don't think it's necessary.
I'm also not sure why you'd want to have the two implementations in parallel.
Given the number of problems there is already, I think it's better to keep it as simple and stupid as possible. 
Two flows mean two pathways for bugs...

Technically the two modules could coexist, since their purposes aren't really the same. 
I'll let you see whether you think it's useful or not and if this PR should be amended about that.

yvaucher: It's a v12 module that pre-existed.

By the way using batches allows the user to follow the progress of the sending.

As for the new state IMO it is there to state the records are being processed, which makes sense to me.

@qgroulard might have more insight on this as he developed those.

qgroulard: It was all about improving the UX by giving feedback to the user.
The batches so that users can follow the progress of the sending.
The new state to get the credit control lines out of the state "To do" asap.

That being said, I can't remember why we did prefer a new state instead of the state "queued".

About not keeping the two implementations in parallel, it looks fair to me, we only use jobs anyway. Although it was quite simple in the v12 module, only a small parameter that doesn't cost much.

yvaucher: @len-foss discussed with @qgroulard , the state seems indeed unecessary, what is important for us though is the batch as it provide a way for the user to quickly look at the mail jobs and see if any fails.

len-foss: @yvaucher so we could have a system parameter and instead of creating a job for each record, create a job by batch of size this parameter? (do you have a suggested value for the default?)

yvaucher: Disabled by default is fine ( =1 ) like you did in your fixup.

account_credit_control_queue_job/models/credit_control_communication.py
```diff
@@ -0,0 +1,23 @@
...
+        try:
+            batch_size = max(1, int(batch_size))
+        except Exception:  # pylint: disable=broad-except
+            batch_size = 1
+        for comms in split_every(batch_size, self.ids, self.browse):
+            if batch_size > 1:
+                desc = _("Sending credit control emails for ids: %s") % comms.ids
+            else:
+                desc = _("Sending credit control email for %s") % comms.partner_id.name
+            comms.with_delay(description=desc)._send_communications_by_email()
```
yvaucher: Sorry for being unclear, by "batch" I meant `queue.job.batch`. It is one job per communication but they are grouped in batches.

https://github.com/OCA/credit-control/blob/12.0/account_credit_control_queue_job/wizard/credit_control_emailer.py#L17-L36

len-foss: All right, done in https://github.com/OCA/credit-control/pull/467/commits/ae9d77b43b9918cd384f8263e3784bc5b2ac3881
I had seen some other custom implementation of the feature use such a batch size for the email sending, so I thought that's what you meant.

qgroulard (approved):
Looks good to me (code review).
Much welcome simplification and improvements.

It would be great to have insight from @sergio-teruel though.

yvaucher (approved):

yvaucher:
@pedrobaeza This is approved and ready to be merged if  you or @sergio-teruel want to have a look before merge please say so.

pedrobaeza (dismissed):
This seems to remove the threaded printing, so @sergio-teruel should check

yvaucher:
@pedrobaeza would it be ok for you if we merge this with a major version? To me the queue job option in a separate module is way more elegent.

pedrobaeza:
@sergio-teruel should tell.

sergio-teruel (approved):

pedrobaeza:
@len-foss please squash commit history a bit removing also the fixups for merging.

len-foss:
@pedrobaeza done

pedrobaeza:
I still see too many commits. For example, there's no sense IMO in having 2 commits about `account_credit_control_queue_job` while you are adding such module here...

len-foss:
How expensive are commits? I left because it also adds another dependency. 
Granted it's part of the same module and everything, but more lines mean more bugs. So having it as an independent commit makes it easier to remove.

pedrobaeza:
For me the architecture has gone too far, depending on another more module `queue_job_batch`. Please follow the KISS principle.

len-foss:
What does this mean? The job batch was added on request of @yvaucher. It's in another module that you don't need to install.

pedrobaeza:
Yes, but it's the replacement of the threaded one, so for having the same feature now 3 modules are needed instead of one. I don't see the sense. You can keep the threaded feature being not activated by default, or put that on a separate module, or have a simple replacement, but not this complicated dependency chain.

len-foss:
How is the threading code KISS?
The problem is, it does not work. It only works if your project as a connection pool 3 times bigger than the number of mails you're going to send.  This has been a massive problem for end customers.

If you want to have such a mechanism in place, do it in a custom module that you install when you know you can. Having an OCA dependency being a problem in OCA modules is a very strange point of view.

pedrobaeza:
Well, OK, I'm not saying that, but I abstain of this PR. Proceed as you wish.

yvaucher:
@len-foss thank you for your reactivity. And also thanks to all others involved for their time.

As it was approved I proceeded with the merge.

I also conclude that if someone needs the threaded version back it should go in a separate module.

queue_job is not an easy dependency to setup but in my experience it's a must have in big projects. And it's more likely that big projects need an async feature. So it's good to have it in a separate module.

len-foss:
Thank you @yvaucher (and @qgroulard ), your help was really appreciated :+1:

PR #470, opened 2025-07-11
[18.0][ADD] account_credit_control_aggregate_level - Aggregate multiple levels in credit lines to make single communications

yvaucher:
Forward port of #22 

In case of recurring invoicing you might have multiple lines with different levels, however you don't want to multiply the communications to the creditor.

This adds the possibility to communicate only the highest due and attach the lower levels to it to not de-multiply the communications.

Compared to `Apply max policy level`, the aim is to keep distinct level values on credit lines and not create them at the same level.

"Auto Process Lower levels" means that you have to care only about the highest level and that other lines for the same partner, policy will be communicated as well.


account_credit_control/models/credit_control_communication.py
```diff
@@ -114,44 +114,60 @@ def _get_credit_lines(
...
+            ("currency_id", "=", currency_id),
+            ("company_id", "=", company_id),
+        ]
+        if level_id:
+            domain.append(("policy_level_id", "=", level_id))
+        return cr_line_obj.search(domain, order="level DESC")
+
+    @api.model
+    def _sql_credit_lines_groups(self):
+        """Create a query to return:
```
len-foss, line 129: Even if it's only called in this one place, you should put the flush inside of it and not in the calling function.

len-foss (commented):
[ADD] is supposed to be for new modules. Here there's a small [REF] refactoring of the module, and mostly [IMP] improvements to it. Which I'm not sure should go into it and not into another module. The rationale for that is that this module is already quite buggy, so I would think keeping it as lightweight as possible would be best.

account_credit_control/models/credit_control_communication.py
yvaucher: I disagree, the flush must be called before the execute.

The rational is that the `_sql_*` method only generate sql content and does nothing else.

len-foss: I misread the Github diff, you're right about this point.

yvaucher:
@len-foss Thanks for your input, I'll modify this PR to create a module `account_credit_control_aggregate_level` to extract this feature.

len-foss:
@yvaucher that's great!
Since you're working on it, it would be good if you also have a check at https://github.com/OCA/credit-control/pull/467
This PR mainly makes sure that it's the postprocess that update the line status based on whether the mail was really sent or not.

There's another that you're unlikely to catch with simple test:
https://github.com/OCA/credit-control/pull/430/
introduced a threading mechanism that makes that a good part of the emails are never sent because they're opened with their own cursor, they get a `PoolError('The Connection Pool Is Full')`, so the run is partly done. The lines have been marked as queued so without modifying data with a server action you have no way to recover from this state.
(so to test it, you need a number of lines that exceeds `ODOO_MAX_HTTP_THREADS`)
I'll push a PR this afternoon.

account_credit_control_aggregate_level/models/credit_control_line.py
```diff
@@ -0,0 +1,143 @@
...
+        Return lines that will receive the same treatment
+        (ie lines of lower level from the same group if grouped).
+        """
+        self.ensure_one()
+        if self.policy_id.aggregate_levels:
+            return self._get_related_lines(level=self.level)
+        else:
+            return self
+
+    def button_schedule_activity(self):
```
len-foss, line 105: Just a question: it seems to have been removed without explanation in https://github.com/OCA/credit-control/commit/e7be29a733f475303407bf16fc4ad24748994a79; should it go back into the module but with a setting to hide the button under certain conditions?

account_credit_control_aggregate_level/tests/test_credit_control_run.py
```diff
@@ -0,0 +1,255 @@
...
+        with invoice_form.invoice_line_ids.new() as invoice_line_form:
+            invoice_line_form.product_id = product
+            invoice_line_form.quantity = 1
+            invoice_line_form.price_unit = 500
+            invoice_line_form.account_id = analytic_account
+            invoice_line_form.tax_ids.clear()
+        cls.invoice = invoice_form.save()
+        cls.invoice.action_post()
+
+    def test_check_run_date(self):
```
len-foss, line 86: It seems this test is copy-pasted from the original module. It does not seem to provide any value, so it would be better to remove it.

account_credit_control_aggregate_level/models/credit_control_line.py
```diff
@@ -0,0 +1,143 @@
...
+        help="`No aggregation` lines are not automatically aggregated "
+        "with other lines.\n"
+        "'Low level' lines are attached under higher level lines.\n"
+        "'Highest level' indicates that all 'Low level' lines for this "
+        "line's partner-policy combination will be attached to it.",
+        default="no",
+        readonly=True,
+    )
+
+    def _update_aggregation(self, exclude_ids=None):
```
len-foss, line 25: Is there a specific reason you can't use a compute, with a trigger on say `communication.line_ids`? The module does really weird things with how relations are managed so it's likely there is one, so a comment would be good to know why precisely (since this is making the unorthodox linking situation worse).

account_credit_control_aggregate_level/tests/test_credit_control_run.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+
+        report_regex = (
+            rf'<p>Policy "<b>{self.policy.name}</b>" has generated <b>'
+            r"\d+ Credit Control Lines.</b><br></p>"
+        )
+        regex_result = re.match(report_regex, control_run.report)
+        self.assertIsNotNone(regex_result)
+
+    def test_generate_credit_lines_with_max_level(self):
```
len-foss, line 140: Same thing; please remove things that are copy-pasted so that the tests actually check for the functional changes that this module brings. Duplicated lines have negative value; in particular these tests are more likely to break due to another module. If the tests are shorter they are more useful as documentation.

account_credit_control_aggregate_level/models/credit_control_line.py
```diff
@@ -0,0 +1,143 @@
...
+                line.write({"aggregation": "no"})
+            if "aggregation" not in values:
+                line._update_aggregation()
+        return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        lines = super().create(vals_list)
+        for line in lines:
+            if line.state == "sent":
```
len-foss, line 53: Is this the case of a force_send, or something else?

yvaucher:
@len-foss could you land an updated review on this one?

len-foss:
@yvaucher It seems I had a pending review from back then where I forgot to push the submit button. I'm not 100% sure they're all valid, but a quick glance suggests these questions still make sense -- so I'm just going to push the button now. Let me know and I can have another look afterwards.

PR #471, opened 2025-07-14
[15.0][FIX] account_credit_control: Use _for_xml_id() to avoid access errors

victoralmau:
Use `_for_xml_id()` to avoid access errors

Please @pedrobaeza can you review it?

@Tecnativa TT57165


pedrobaeza (approved):
Obvious patch, so fast-tracking it:

/ocabot merge patch

PR #472, opened 2025-07-14
[16.0][FIX] account_credit_control: Use _for_xml_id() to avoid access errors

victoralmau:
FWP from 15.0: https://github.com/OCA/credit-control/pull/471

Use `_for_xml_id()` to avoid access errors

Please @pedrobaeza can you review it?

@Tecnativa TT57165

PR #473, opened 2025-07-15
[17.0][FIX] account_credit_control: Use _for_xml_id() to avoid access errors

victoralmau:
FWP from 16.0: https://github.com/OCA/credit-control/pull/472

Use `_for_xml_id()` to avoid access errors

Please @pedrobaeza can you review it?

@Tecnativa TT57165


pedrobaeza (approved):
Fast-tracking as obvious patch:

/ocabot merge patch

PR #481, opened 2025-07-22
[18.0][FIX] account_financial_risk: Fix access rights on credit_limit

tobiaszehntner:
When users without accounting access are trying to access a Sale Order, they have an access error on the partner field credit_limit (standard Odoo field that is limited to Account groups), which is used when computing the SO's risk_info. This is fixed by adding the group financial_risk_user to the field. Now the user can access the SO, and sees the risk_info if they have the financial_risk_user right, otherwise it is not displayed, with no access error.

### DIFF
--- a/account_financial_risk/models/res_partner.py
+++ b/account_financial_risk/models/res_partner.py
@@ -198,20 +198,19 @@ class ResPartner(models.Model):
 
     @api.depends("credit_limit")
     def _compute_date_credit_limit(self):
-        self.filtered(lambda x: x.credit_limit == 0.00).date_credit_limit = False
+        self.filtered(lambda x: x.sudo().credit_limit == 0.00).date_credit_limit = False
         self.filtered(
-            lambda x: x.credit_limit != 0.00
+            lambda x: x.sudo().credit_limit != 0.00
         ).date_credit_limit = datetime.today()
 
     @api.depends("credit_limit", "risk_total")
     def _compute_risk_remaining(self):
         for record in self:
-            record.risk_remaining_value = record.credit_limit - record.risk_total
-            if record.credit_limit:
+            credit_limit = record.sudo().credit_limit
+            record.risk_remaining_value = credit_limit - record.risk_total
+            if credit_limit:
                 record.risk_remaining_percentage = round(
-                    100
-                    * (record.credit_limit - record.risk_total)
-                    / record.credit_limit,
+                    100 * (credit_limit - record.risk_total) / credit_limit,
                     2,
                 )
             else:
@@ -481,7 +480,7 @@ class ResPartner(models.Model):
             credit_limit = partner.sudo().credit_limit
             if credit_limit and amount > credit_limit:
                 risk_exception = True
-                amount_exceeded = amount - partner.credit_limit
+                amount_exceeded = amount - credit_limit
             partner.risk_total = amount
             partner.risk_amount_exceeded = amount_exceeded
             partner.risk_exception = risk_exception
--- a/account_financial_risk/models/account_invoice.py
+++ b/account_financial_risk/models/account_invoice.py
@@ -54,7 +54,7 @@ class AccountMove(models.Model):
         elif not partner.risk_invoice_draft_include and (
             partner.risk_invoice_open_include
             and (partner.risk_total + self.risk_amount_total_currency)
-            > partner.credit_limit
+            > partner.sudo().credit_limit
         ):
             exception_msg = self.env._("This invoice exceeds the financial risk.\n")
         return exception_msg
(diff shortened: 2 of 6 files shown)


acsonefho (approved):

pedrobaeza:
Wasn't this fixed at #465 ?

tobiaszehntner:
> Wasn't this fixed at #465 ?

No my error came from [this line](https://github.com/OCA/credit-control/blob/f1d0363c664d4ef94a630b31061e54c55a6153d3/sale_financial_risk_info/models/sale.py#L27). But there are plenty of places where we access `partner.credit_limit` in this repository. Since it's an Odoo field that is re-purposed for the risk computation, I think it makes sense to add the "Financial Risk User" right to that field. Then we could remove the sudo in the other function?

pedrobaeza:
But the solution should be the same: access it with `.sudo`.

tobiaszehntner:
> But the solution should be the same: access it with `.sudo`.

I count 16 places where we access partner.credit_limit. And the risk group is for "Read only access to financial risk info". Why is it not appropriate to add that group to the field?

If we use sudo each time, don't we just increase the risk to re-create that bug with the next development accessing the field?

pedrobaeza:
Yes, all that 16 parts should be accessed with `sudo`. Changing groups is not the suitable solution. And you are also hardcoding a full list of groups, overriding any other thing, that may be not correct. It's the same as when Odoo removes read access over ir.act_window. All the places (and they were a lot, required a sudo to access it).

carlosdauden (changes requested):
The aim of these modules is to control the risk for all users, especially for those with fewer permissions, so limiting access to the credit_limit field is not the best solution.

tobiaszehntner:
> The aim of these modules is to control the risk for all users, especially for those with fewer permissions, so limiting access to the credit_limit field is not the best solution.

Exactly my point, I'm *extending* the access (adding the financial risk group [to the ones Odoo adds](https://github.com/odoo/odoo/blob/9ec4e0d3c5f1aaff0b36dc5def1e69b8285c3be1/addons/account/models/partner.py#L522)). 

But I agree that it's overwriting the groups for anyone after us as well, and if Odoo changes their groups in the future, it will unnecessarily add the Accounting groups as well, so `sudo` it is.

LauraCForgeFlow:
I also came across an issue with this. What about for views, such as in this line [here](https://github.com/OCA/credit-control/blob/f897758d93a340d257c7815cdf2833a0540d000c/account_financial_risk/views/res_partner_view.xml#L216)? Users with Financial Risk permissions but without Accounting permissions will get an access error when trying to display this field. 

Should we restrict it on the view to users with Accounting permissions? But then the Financial Risk users would be missing out on this information. I'm not sure how to approach it. Thanks in advance!!

AaronHForgeFlow (commented):
I am not sure that adding sudo everywhere is a good option. Isn't it better to modify the record rules to provide read access to the needed models?

That would also prevent errors in views with restricted fields like one mentioned in the previous comment: https://github.com/OCA/credit-control/pull/481#issuecomment-3112531467

pedrobaeza:
@AaronHForgeFlow for this specific case, it is. It's not a question of record rules, but specific field protection that can't be bypassed through that tool.

@tobiaszehntner please rebase and resolve conflicts.

tobiaszehntner:
> @tobiaszehntner please rebase and resolve conflicts.

done

pedrobaeza (approved):
Code review.

@LauraCForgeFlow about the view, you are right that there can be a mismatch. What can be done is to have a related sudo field that is the one shown in that tab. What do you think?

LauraCForgeFlow:
Good idea @pedrobaeza! I have implemented it in this PR: #494.

carlosdauden (approved):

account_financial_risk/models/res_partner.py
```diff
@@ -198,20 +198,19 @@ def _compute_portal_show_financial_risk_visible(self):
 
     @api.depends("credit_limit")
     def _compute_date_credit_limit(self):
-        self.filtered(lambda x: x.credit_limit == 0.00).date_credit_limit = False
+        self.filtered(lambda x: x.sudo().credit_limit == 0.00).date_credit_limit = False
```
carlosdauden, line 201: Why do these fields not have `compute_sudo=True` set like the rest of the fields?

account_financial_risk/models/res_partner.py
```diff
@@ -198,20 +198,19 @@ def _compute_portal_show_financial_risk_visible(self):
...
-            lambda x: x.credit_limit != 0.00
+            lambda x: x.sudo().credit_limit != 0.00
         ).date_credit_limit = datetime.today()
 
     @api.depends("credit_limit", "risk_total")
     def _compute_risk_remaining(self):
         for record in self:
-            record.risk_remaining_value = record.credit_limit - record.risk_total
-            if record.credit_limit:
+            credit_limit = record.sudo().credit_limit
```
carlosdauden, line 209: Why do these fields not have `compute_sudo=True` set like the rest of the fields?

https://github.com/OCA/credit-control/blob/a0ce430f5bb4be30352174c8835a293d6c2d89f4/account_financial_risk/models/res_partner.py#L29-L31

https://github.com/OCA/credit-control/blob/a0ce430f5bb4be30352174c8835a293d6c2d89f4/account_financial_risk/models/res_partner.py#L48-L50

...

PR #487, opened 2025-08-05
[16.0][IMP] account_financial_risk: Display credit information in portal view

pilarvargas-tecnativa:
@Tecnativa TT57363

@pedrobaeza @victoralmau please review


pedrobaeza:
Screenshot of the change?

pilarvargas-tecnativa:
> Screenshot of the change?

image

It has also been documented in readme.

account_financial_risk/models/res_config.py
```diff
@@ -13,3 +13,8 @@ class AccountConfigSettings(models.TransientModel):
     allow_overrisk_invoice_validation = fields.Boolean(
         related="company_id.allow_overrisk_invoice_validation", readonly=False
     )
+    portal_show_financial_risk = fields.Boolean(
```
pedrobaeza, line 16: It would be interesting also to have the option by partner (apart from this one), enabled by default, but invisible if this parameter is not set.

account_financial_risk/views/portal_templates.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2025 Tecnativa - Pilar Varga
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="portal_layout" inherit_id="portal.portal_layout">
+        <xpath expr="//div[hasclass('o_portal_my_security')]" position="after">
+            <div
+                t-if="request.env['ir.config_parameter'].sudo().get_param('account_financial_risk.portal_show_financial_risk') and user_id.partner_id.credit_limit != 0"
```
victoralmau:
```suggestion
                t-if="request.env['ir.config_parameter'].sudo().get_param('account_financial_risk.portal_show_financial_risk') and user_id.partner_id.credit_limit != 0"
```
I think this `t-if` should be simplified and a variable defined in the `_prepare_portal_layout_values()` method (I think that would be the right one, but I'm not entirely sure) with the appropriate value. That way, if necessary, the value could be changed and this could be “_hidden_” in some cases.

account_financial_risk/views/portal_templates.xml
```diff
@@ -0,0 +1,80 @@
...
+                class="o_portal_my_financial_risk mt-3"
+            >
+                <h4>Credit information</h4>
+                <hr class="mt-1 mb-1" />
+                <ul class="list-group">
+                    <li class="list-group-item bg-transparent border-0"><strong
+                        >Credit limit </strong><span
+                            t-field="user_id.partner_id.credit_limit"
```
victoralmau: Are you sure this isn't a problem for a “_basic_” user?

I'm not 100% sure, but the field has defined groups https://github.com/odoo/odoo/blob/d024933cee7b99cfce0823bd2a2828dcbdc60184/addons/account/models/partner.py#L477 and you probably need to use `sudo()` to avoid any potential problems. Perhaps something similar happens in other fields.

victoralmau (changes requested):

account_financial_risk/views/portal_templates.xml
pilarvargas-tecnativa: This field only has groups.

account_financial_risk/views/portal_templates.xml
pilarvargas-tecnativa: It has been changed to check everything at the contact level, no further code should be added.
Does that look correct to you?

account_financial_risk/models/res_config.py
pilarvargas-tecnativa: done

account_financial_risk/readme/USAGE.rst
```diff
@@ -9,3 +9,10 @@ To use this module, you need to:
 #. Test the restriction trying to create an invoice for the partner for an
    amount higher of the limit you have set.
 #. Return to Customer *Financial Risk* tab and click in amount to view origin.
+
+(Optional) To display financial risk information in the customer portal,
```
pedrobaeza: This should go into `CONFIGURE.rst`.

pedrobaeza (changes requested):
For company contacts, put the check "Show financial risk in portal" inside the "Financial risk" page. For children contacts, you should put it in the popup window:

imagen

And call it "Show credit information in portal".

And the children portal is not correct:

imagen

comparing with the parent one:

imagen

pilarvargas-tecnativa:
> For company contacts, put the check "Show financial risk in portal" inside the "Financial risk" page. For children contacts, you should put it in the popup window:
> And call it "Show credit information in portal".

image

> And the children portal is not correct:

image

pedrobaeza:
Why have you renamed the internal variables? I was talking only about the label.

account_financial_risk/models/res_partner.py
```diff
@@ -174,8 +174,8 @@ class ResPartner(models.Model):
         help="If enabled, this partner will see their financial risk in the portal, "
         "provided the global setting is also enabled.",
     )
-    portal_show_financial_risk_enabled = fields.Boolean(
-        compute="_compute_portal_show_financial_risk_enabled",
+    portal_show_credit_info_enabled = fields.Boolean(
```
pedrobaeza: Talking about variable names, why `show` and `enabled` in the same variable? only one of them is neeeded, not both.

pilarvargas-tecnativa: To distinguish the res.config.settings option ‘portal_show_credit_info’ from res.partner, which is calculated to check whether it is active in the configuration for visibility.

pedrobaeza: Then use `visible` instead of `enabled`, isn't it?

pilarvargas-tecnativa:
> Why have you renamed the internal variables? I was talking only about the label.

So that the variable name is consistent with its string. The real purpose is to display the credit information and nothing else, so I thought the variable name should also be in accordance with that.

account_financial_risk/models/res_partner.py
pilarvargas-tecnativa: Done

pedrobaeza:
One thing is the variable (technical name), and other the label. Check for example when you have a variable at technical level called `move_line_ids`, but the label is "Journal items", not "Move lines", and you don't name the variable `journal_item_ids`. But well, this is only for general knowledge information.

pedrobaeza (approved):
Tested on runboat.

pilarvargas-tecnativa:
> One thing is the variable (technical name), and other the label. Check for example when you have a variable at technical level called `move_line_ids`, but the label is "Journal items", not "Move lines", and you don't name the variable `journal_item_ids`. But well, this is only for general knowledge information.

Done

account_financial_risk/views/portal_templates.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2025 Tecnativa - Pilar Varga
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="portal_layout" inherit_id="portal.portal_layout">
+        <xpath expr="//div[hasclass('o_portal_my_security')]" position="after">
+            <div
+                t-if="user_id.partner_id.portal_show_financial_risk_visible and user_id.partner_id.show_financial_risk_in_portal and user_id.partner_id.sudo().credit_limit != 0"
```
victoralmau:
```suggestion
                t-if="user_id.partner_id.portal_show_financial_risk_visible and user_id.partner_id.show_financial_risk_in_portal and user_id.partner_id.sudo().commercial_partner_id.credit_limit != 0"
```
It might be useful to define a variable (`partner_financial_risk = user_id.partner_id.sudo().commercial_partner_id`) and use it everywhere.

victoralmau (changes requested):

account_financial_risk/views/portal_templates.xml
pilarvargas-tecnativa: Good tip! thanks!

victoralmau (approved):
Functional review OK

account_financial_risk/models/res_partner.py
```diff
@@ -168,6 +168,32 @@ class ResPartner(models.Model):
...
+
+    @api.depends()
+    def _compute_portal_show_financial_risk_visible(self):
+        global_enabled = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("account_financial_risk.portal_show_financial_risk")
+        )
+        for partner in self:
+            partner.portal_show_financial_risk_visible = global_enabled == "True"
```
carlosdauden: Loop is needed?

```suggestion
        self.portal_show_financial_risk_visible = global_enabled == "True"
```

carlosdauden (changes requested):

account_financial_risk/models/res_partner.py
pilarvargas-tecnativa: thanks

account_financial_risk/models/res_partner.py
```diff
@@ -168,6 +168,31 @@ class ResPartner(models.Model):
...
+        ]
+
+    @api.depends()
+    def _compute_portal_show_financial_risk_visible(self):
+        global_enabled = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("account_financial_risk.portal_show_financial_risk")
+        )
+        self.portal_show_financial_risk_visible = global_enabled == "True"
```
pedrobaeza: I think you should use `str2bool` method for allowing "True", "1", etc.

pilarvargas-tecnativa: done

carlosdauden (approved):

PR #488, opened 2025-08-06
[18.0][IMP] account_credit_control - multi-channel policies

yvaucher:
(superseeded by #522 )

Adds the possibility to send reminders via multiple channels for the same level.

Sometimes you want to send reminders on multiple channels at the same time for instance send an email as well as sending a letter on high policy level.

This adds a checkbox for each channel.

Adding a new channel must be done by adding the following fields:

- `channel_<name>` on both `credit.control.line` and `credit.control.policy` 
- `<name>_sent` on `credit.control.line`

Line is updated to `sent` only after all selected channels have been processed. It is still possible to force the manually the state to sent to settle the line.

Policy level:

image

Credit control line:

image

TODO:

- [x] Rebase on top of #467 
- [x] Migration script to set channels

### DIFF
--- a/account_credit_control/models/credit_control_line.py
+++ b/account_credit_control/models/credit_control_line.py
@@ -6,8 +6,6 @@
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
-from .credit_control_policy import CHANNEL_LIST
-
 
 class CreditControlLine(models.Model):
     """A credit control line describes an amount due by a customer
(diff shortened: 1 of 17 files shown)


account_credit_control/models/credit_control_line.py
```diff
@@ -62,11 +60,15 @@ class CreditControlLine(models.Model):
         "Draft and ignored lines will be "
         "generated again on the next run.",
     )
-    channel = fields.Selection(
-        selection=CHANNEL_LIST,
-        required=True,
```
qgroulard, line 67: We lose this `required`, I guess we should add a check to make sure that at least one channel field is `True`.

qgroulard (commented):
Great addition :+1: 

Multi-channel seems to be a really common need to me (already supported by Odoo Follow-Up btw).
I like this simple solution.

yvaucher:
Fixed tests by taking in account `account_credit_control_queue_job` and ensuring jobs if any are run.

baimont (approved):
nice pr, a small fix to make tests pass

qgroulard (approved):

yvaucher:
Tests to be fixed

PR #491, opened 2025-08-18
[16.0][ADD] website_sale_financial_risk: New module

pilarvargas-tecnativa:
@Tecnativa TT57364

@pedrobaeza @carlos-lopez-tecnativa please review


pedrobaeza (approved):
Change the text "This order will be registered as on credit." to "The order will be confirmed on credit with no need of instant payment, as you still have available credit.".

Please do as well the method to get the risk fields and expand the other module for showing parametrically the fields

pilarvargas-tecnativa:
> Please do as well the method to get the risk fields and expand the other module for showing parametrically the fields

I have added the inheritable template to the base module, and as soon as it is merged, I will add the necessary fields from the missing module. Simply by calling that template, all the fields will appear.

https://github.com/OCA/credit-control/pull/491/commits/b31b8d28487c322cc7375b4d836fee03f7319536

pedrobaeza:
That's not the strategy I'm telling you. What I'm saying is:

- In `account_financial_risk`, add a method in `res.partner` called `_get_financial_risk_fields`, returning the fields already defined there.
- In dependent modules, override and add the fields of that module.
- On the base module, for the portal template, do a loop over the fields returned by the method to fill each line, getting the description from the field label.

pilarvargas-tecnativa:
> That's not the strategy I'm telling you. What I'm saying is:

I understand your approach, but with that method, you would have to check whether the field is marked or not to display it. Not all fields are displayed, and that type of logic is more suited to a controller, in my opinion. That is why I opted to create an inheritable view where the fields can be added with their corresponding t-if, and the view can be added simply with the t-call. Both approaches can be combined with t-foreach, but as I said, you would have to check field by field whether or not to add them to the fields to be displayed, as long as they are of the monetary type, in order to have a common widget in that t-foreach.

image

pedrobaeza:
Then the method should return a list of tuples with 2 elements: the field of the check, and the amount field, but adding a new parameter in subsequent modules shouldn't be so complicated, as if not, for sure they will forget to add the field to the portal template (or whatever other template we need - like mail template -). It's a non scalable approach. Please go with mine.

carlos-lopez-tecnativa (approved):

account_financial_risk/models/res_partner.py
```diff
@@ -567,3 +567,52 @@ def open_risk_pivot_info(self):
...
+        self.ensure_one()
+        return [
+            (
+                self.risk_invoice_draft_include,
+                formatLang(
+                    self.env,
+                    self.risk_invoice_draft,
+                    currency_obj=self.risk_currency_id,
+                ),
+                _("Draft Invoices"),
```
pedrobaeza: Why hard-coding strings?

account_financial_risk/models/res_partner.py
```diff
@@ -567,3 +567,52 @@ def open_risk_pivot_info(self):
...
             "domain": domain,
         }
+
+    def _get_financial_risk_lines(self):
+        # Returns [(flag, value, label), ...] already evaluated for the partner itself.
+        self.ensure_one()
+        return [
+            (
+                self.risk_invoice_draft_include,
+                formatLang(
```
pedrobaeza: I don't think you should pass the values, but the field names, or at least, the floats.

account_financial_risk/models/res_partner.py
pilarvargas-tecnativa: To reduce the length of the strings and avoid the word ´Total´
https://github.com/OCA/credit-control/blob/16.0/account_financial_risk/models/res_partner.py#L32

pedrobaeza: Yeah, but that means more terms to translate, and very similar. I don't see at all to call it "Total Draft Invoices", or to rename the field label.

account_financial_risk/models/res_partner.py
pilarvargas-tecnativa: t-field does not support dynamic expressions, so passing the field name would not return the value.
With t-esc or t-out, you could obtain the flat value but without the currency symbol. One possible solution would be to include the currency symbol separately, but in what position? Before or after the value? Because it depends on the currency, and this is done automatically by the widget and cannot be applied in this case.

pedrobaeza: Yes, it does, with `t-esc-options`.

pilarvargas-tecnativa: t-options applied to t-esc only works in QWeb reports (PDF/HTML generated by report_qweb) and not in QWeb views such as those in the portal or web views, in which case it would apply to t-field.

I have been checking to see if there is anything new on this topic, but I cannot find any documentation or examples on it.

pedrobaeza: Have you tried with `t-esc-options`, not `t-options`?

pilarvargas-tecnativa: done

account_financial_risk/models/res_partner.py
```diff
@@ -567,3 +567,79 @@ def open_risk_pivot_info(self):
...
+        #             currency_obj=self.risk_currency_id,
+        #         ),
+        #         self._fields["risk_invoice_draft"].string,
+        #     ),
+        #     (
+        #         self.risk_invoice_open_include,
+        #         formatLang(
+        #             self.env, self.risk_invoice_open, currency_obj=self.risk_currency_id
+        #         ),
+        #         self._fields["risk_invoice_open"].string,
```
yvaucher: There are some left over comments, is it intended?

pilarvargas-tecnativa: Residual code removed, thanks

yvaucher (approved):

pilarvargas-tecnativa:
@yvaucher The PR has closed without merging.

yvaucher:
@pilarvargas-tecnativa While the state of this PR is indeed strangely wrong (mentionned as closed instead of merged) your commits are now in the branch 16.0

https://github.com/OCA/credit-control/commits/16.0/

It's not the ones on the top because of the ordering by date but you can still see them on the first page.

image


Thanks for your contribution!

pilarvargas-tecnativa:
The error was caused by the last push after the command had already been executed. Although there were no new changes, it caused the PR status to be incorrect.

PR #494, opened 2025-08-25
[18.0][FIX] account_financial_risk: Fix access rights for credit_limit on views

LauraCForgeFlow:
As agreed on https://github.com/OCA/credit-control/pull/481#pullrequestreview-3144443698, a sudo field related to the original `credit_limit` has been created, with its same specifications, and it is the one show on the actual view. This is to avoid permissions problems as well as keep all the relevant information visible.


AaronHForgeFlow (approved):
:laughing:  weird but ok!

pedrobaeza (approved):
@carlosdauden do you agree?

account_financial_risk/migrations/18.0.1.1.1/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+
+def migrate(cr, version):
+    cr.execute(
```
pedrobaeza: I think you can do `from ....hooks import pre_init_hook and call it from here for not duplicating code.

account_financial_risk/models/res_partner.py
```diff
@@ -135,7 +135,18 @@ class ResPartner(models.Model):
...
-    credit_limit = fields.Float(tracking=True)
+    credit_limit = fields.Float(string="Credit Limit (Original)")
+    credit_limit_sudo = fields.Float(
+        string="Credit Limit",
+        help="Credit limit specific to this partner.",
+        company_dependent=True,
+        copy=False,
+        readonly=False,
+        store=True,
+        tracking=True,
```
pedrobaeza, line 148: It's correct to track this? Isn't duplicated?

LauraCForgeFlow: I removed the tracking from the original `credit_limit` field to avoid tracking it two times.

pedrobaeza: But I don't think there's a field access problem at that level, isn't it? Or is it hidden?

LauraCForgeFlow: I have tested it locally and the problem is for users who don't have access to the original `credit_limit` field. Because for them, the changes in the original field are not displayed in the chatter, losing that specific information. If we do it with the `credit_limit_sudo` field, everyone can see the tracking.

carlosdauden:
I don't quite understand adding the credit_limit_sudo field, which is going to be available for all groups...
Isn't that practically the same as removing the groups from the original field?

I think this can be solved in two different ways:
 - Remove the groups from that field when installing this module.
 - Use sudo when querying that field by code. (Another option is to create a _get_credit_limit method to use instead of sudo). At the view level it should not be available to users who do not belong to the specified groups.

carlosdauden (changes requested):
I prefer this option: #481 

If you consider it essential, you can propose adding the group_account_financial_risk_manager group to the credit_limit field in that PR, but normally whoever manages financial risk has accounting permissions.

pedrobaeza:
I think this solution can get legit adding to the field the risk manager group, as we want that group to access and set this value without accounting permissions.

LauraCForgeFlow:
So I should add the field in the original `credit_limit` field? Or to the new `sudo` field?

I initially though about adding it to the original `credit_limit` field, but this [comment](https://github.com/OCA/credit-control/pull/481#issuecomment-3112333323) in the mentioned PR suggested this was not a good practice. Is it the best solution for this specific case, though?

pedrobaeza:
I mean in the new one for not breaking inheritances.

account_financial_risk/models/res_partner.py
```diff
@@ -135,7 +135,21 @@ class ResPartner(models.Model):
     )
     credit_policy = fields.Char()
     risk_allow_edit = fields.Boolean(compute="_compute_risk_allow_edit")
-    credit_limit = fields.Float(tracking=True)
+    credit_limit = fields.Float(string="Credit Limit (Original)")
```
pedrobaeza, line 138: According what you said:

```suggestion
    credit_limit = fields.Float(string="Credit Limit (Original)", tracking=False)
```

pedrobaeza: Oh, I see the point... forget it.

pedrobaeza:
See alternative in #496

PR #496, opened 2025-08-28
[18.0][IMP] account_financial_risk: Add risk group to credit_limit field

carlosdauden:
I'm not very keen on adding code when there's no real-world need for it, but if we're going to do something ugly, I prefer this approach to the one proposed in #494. 

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
@LauraCForgeFlow this seems a good alternative

LauraCForgeFlow:
I agree! This is not as ugly as my original proposal. 

Closing that PR, thank you both! :)

PR #511, opened 2025-10-22
[17.0][MIG] base_upflow: Migration to 17.0

Koudja:


pedrobaeza:
This seems overkill to split the module into 2, and putting this one here in OCA/credit-control, which is not something related, as you can't do anything in Odoo with this base module.

PR #522, opened 2025-11-14
[18.0][IMP] account_credit_control - multi-channel policies

yvaucher:
(superseeds #488)

Adds the possibility to send reminders via multiple channels for the same level.

Sometimes you want to send reminders on multiple channels at the same time for instance send an email as well as sending a letter on high policy level.

This adds a checkbox for each channel.

Adding a new channel must be done by adding the following fields:

- `channel_<name>` on both `credit.control.line` and `credit.control.policy` 
- `<name>_sent` on `credit.control.line`

Line is updated to `sent` only after all selected channels have been processed. It is still possible to force the manually the state to sent to settle the line.

Policy level:

image

Credit control line:

image

TODO:

- [x] Rebase on top of #467 
- [x] Migration script to set channels

### DIFF
--- a/account_credit_control/__manifest__.py
+++ b/account_credit_control/__manifest__.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Account Credit Control",
-    "version": "18.0.2.0.0",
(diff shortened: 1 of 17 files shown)


bosd:
:eyes: this one, for :green_circle: tests

PR #527, opened 2025-12-02
[18.0][FIX] account_credit_control: Freeze test execution date to avoid time-dependent failures

pilarvargas-tecnativa:
@Tecnativa 

@pedrobaeza @carlos-lopez-tecnativa please review

### DIFF
--- a/account_credit_control/tests/test_credit_control_run.py
+++ b/account_credit_control/tests/test_credit_control_run.py
@@ -6,6 +6,7 @@ import re
 from datetime import datetime
 
 from dateutil import relativedelta
+from freezegun import freeze_time
 
 from odoo import fields
 from odoo.exceptions import AccessError, UserError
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But you should freeze it also in the test methods.

pilarvargas-tecnativa:
> But you should freeze it also in the test methods.

In this case, freeze_time can be applied at the class level since, as it is a post_install test, it is not part of the minimum installation battery and the entire suite shares a single consistent time scenario.

pedrobaeza:
But I think there's a problem in the library that provokes the freeze, and that's why you should put it on each method instead. @victoralmau found this problem.

pilarvargas-tecnativa:
> But I think there's a problem in the library that provokes the freeze, and that's why you should put it on each method instead. @victoralmau found this problem.

In that case, it is better to freeze only the test method that needs it to avoid greater harm by generalizing.

pedrobaeza (approved):
Fast-tracking this patch for avoiding glitches on extreme dates.

/ocabot merge patch

PR #531, opened 2025-12-10
[18.0][MIG] website_sale_financial_risk: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT59371

@victoralmau @eduezerouali-tecnativa please review


website_sale_financial_risk/models/payment_transaction.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, models
```
eduezerouali-tecnativa:
```suggestion
from odoo import models
```

website_sale_financial_risk/models/payment_transaction.py
```diff
@@ -0,0 +1,42 @@
...
+        tx = super()._get_tx_from_notification_data(provider_code, notification_data)
+        if provider_code != "credit" or len(tx) == 1:
+            return tx
+        reference = notification_data.get("reference")
+        tx = self.search(
+            [("reference", "=", reference), ("provider_code", "=", "credit")]
+        )
+        if not tx:
+            raise ValidationError(
+                _("On Credit: No transaction found matching reference %s.", reference)
```
eduezerouali-tecnativa:
```suggestion
                self.env._("On Credit: No transaction found matching reference %s.", reference)
```

eduezerouali-tecnativa (approved):
LGTM + tested on runboat. Just that minor thing.

website_sale_financial_risk/models/payment_transaction.py
pilarvargas-tecnativa: thanks!

victoralmau (approved):
Code and functional review OK.

It would be interesting to add a specific `payment.method` as demo data and link it to the corresponding `payment.provider` with the `payment_method_ids` field.

clementmbr:
Hi @pilarvargas-tecnativa  or @victoralmau , my question may be silly, but how do you test or use this module if there is no `payment.method` linked to the "Credit" provider ?
I thought a `payment.method` was mandatory to make a payment on the website.

How are you supposed to link the Credit provider to the `account.payment.method` created in the data and is it possible to use it to make a payment on the website?

pilarvargas-tecnativa:
> How are you supposed to link the Credit provider to the `account.payment.method` created in the data and is it possible to use it to make a payment on the website?

Hello! Setting up a provider with the current changes can be confusing if you're used to previous versions.

That's right. To make online payments, you need a billing method associated with the provider. 

This data isn't in the module, but you can create and configure it as long as the provider supports it. In this case, it would be a “symbolic” payment method, such as a bank transfer, since the system will process the payment using the available credit.

I hope this clears up your question.

Best regards

PR #532, opened 2025-12-11
[16.0][IMP] account_invoice_overdue_reminder: add compatibility with OCA module mail_composer_cc_bcc

alexis-via:


alexis-via:
It doesn't work because unfortunately the OCA module mail_composer_cc_bcc only works with mail composer, not for regular email from queue job.

PR #536, opened 2025-12-23
[18.0][FIX] sale_financial_risk: avoid writing on commercial_partner during compute on partner

clementmbr:
Currently Odoo is writing on a commercial_partner when doing `_compute_risk_sale_order()` on a partner (two different records).

It is against the expected behavior of a non-stored readonly computed field and can create **security problems** as the commercial_partner record is not necessary accessible by the Odoo user who is accessing the current partner's record.

cc @sebastienbeau

### DIFF
--- a/sale_financial_risk/models/res_partner.py
+++ b/sale_financial_risk/models/res_partner.py
@@ -27,7 +27,7 @@ class ResPartner(models.Model):
         risk_states = self.env["sale.order"]._get_risk_states()
         return self._get_risk_company_domain() + [
             ("state", "in", risk_states),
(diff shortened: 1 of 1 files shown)


ivantodorovich:
Why is it writing on commercial_partner_id upon read? the field is stored

IMO previous code was ok

cyrilmanuel (approved):

pedrobaeza (changes requested):
This is needed since this change:

https://github.com/akretion/credit-control/commit/317c47b100a69c14ad7ba2f5b8a52f329105cc91

and being non stored, it's not a problem.

Please indicate which is the specific problem you want to fix.

PR #545, opened 2026-02-20
[17.0][FIX] account_credit_control: Apply sudo() to avoid access error

victoralmau:
Apply `sudo()` to avoid access error + Add `sale_financial_risk` test use case

@Tecnativa TT61117


pedrobaeza (approved):

flotho (approved):
Easy code review LGTM
Thanks

PR #547, opened 2026-02-20
[18.0][FIX] account_credit_control: Apply sudo() to avoid access error

victoralmau:
FWP from 17.0: https://github.com/OCA/credit-control/pull/545

Apply `sudo()` to avoid access error + Add `sale_financial_risk` test use case

@Tecnativa TT61117


pedrobaeza (approved):
/ocabot merge patch

PR #559, opened 2026-04-26
[19.0][ADD] account_financial_risk_global: new module

gdgellatly:
## Summary

- New module `account_financial_risk_global` that evaluates partner financial risk across all companies
- The `credit_limit` field on `res.partner` is global, but `account_financial_risk` restricts risk queries to the current user's companies, creating a gap where a partner can exceed their limit across companies without triggering an exception from any single company's view
- This module removes the company filter (`_get_risk_company_domain` returns `Domain.TRUE`) and runs the risk computation with `sudo()` so that totals always reflect the partner's full exposure

## Technical details

- Overrides `_get_risk_company_domain()` to return `Domain.TRUE` (include all companies)
- Overrides `_compute_risk_account_amount()` to run via `sudo()` so cross-company journal items are visible
- Amounts from other companies' receivable accounts are classified under `risk_account_amount` / `risk_account_amount_unpaid` (not `risk_invoice_open`) since `property_account_receivable_id` is company-specific — but they still count toward `risk_total`
- No configuration required — installing the module is sufficient

## Test plan

- [x] 6 unit tests covering multi-company scenarios
- [x] `test_risk_domain_is_global` — domain returns `Domain.TRUE`
- [x] `test_risk_totals_span_companies` — draft invoices in 2 companies sum correctly
- [x] `test_exception_triggered_globally` — credit limit exceeded with combined amounts
- [x] `test_no_exception_within_limit` — no false positive within limit
- [x] `test_posted_invoices_span_companies` — posted invoices across companies count in risk total
- [x] `test_single_company_user_still_sees_global_risk` — user with single-company access sees full global risk
- [x] All pre-commit hooks pass (ruff, pylint_odoo, prettier)
- [x] Clean install + test-enable on fresh database


carlosdauden:
I don’t think it’s necessary to create a new module for this; simply add a system parameter to the base module and propose the following change:

```
    @api.model
    def _get_risk_company_domain(self):
        if self.env['ir.config_parameter'].sudo().get_param('account_financial_risk.all_companies_risk'):
            return Domain.TRUE
        else:
            return [("company_id", "in", self.env.companies.ids)]

```
If you like, you can add an option in the settings to improve the user experience.

An even more powerful option would be to maintain a company-wide list of the companies for which risk is calculated.

```
    @api.model
    def _get_risk_company_domain(self):
        return [("company_id", "in", self.env.company.risk_related_companies.ids or self.env.companies.ids)]

```

carlosdauden:
ping @sergio-teruel

PR #561, opened 2026-05-07
[17.0][ADD] sale_financial_risk_exclude_order_type

atgalvez08:
@BinhexTeam

## Description

This module extends `sale_financial_risk` to support
financial risk behavior configuration directly on
`sale.order.type`.

It allows configuring sale order types to:

- exclude orders from financial risk computation
- bypass financial risk blocking validation

The implementation intentionally keeps the design
simple and business-oriented by integrating the
configuration directly into sale order types.

## Features

- Exclude order types from risk computation
- Allow blocking bypass by order type
- Independent control of computation and blocking
- Multi-company compatible behavior
- Seamless integration with `sale_order_type`

## Included components

- `sale.order.type` extensions
- Risk computation domain extension
- Financial risk blocking override
- Automated test suite

## Design considerations

The implementation intentionally avoids introducing
additional configuration models.

The module:

- keeps configuration close to business semantics
- reduces maintenance complexity
- avoids duplicating framework logic
- preserves existing `sale_financial_risk` behavior
- follows a lightweight extension approach

## Technical details

- Extends `sale.order.type`
- Overrides `evaluate_risk_message`
- Extends `_get_risk_sale_order_domain`
- Supports optional blocking bypass
- Supports optional exclusion from risk computation

## Dependencies

- sale_financial_risk
- sale_order_type

## Testing

Includes automated tests for:

- blocking bypass behavior
- risk computation domain generation
- separation of concerns
- neutral order types
- multi-company behavior
- domain integrity

## Checklist

- [x] Module follows OCA structure
- [x] README aligned with implementation
- [x] Tests included
- [x] No duplicated framework logic
- [x] License headers included


sale_financial_risk_exclude_order_type/tests/test_sale_financial_risk_exclude_order_type.py
```diff
@@ -0,0 +1,289 @@
...
+        super().setUpClass()
+        cls.company = cls.env.company
+        cls.partner = cls.env["res.partner"].create(
+            {
+                "name": "Risk Customer",
+                "customer_rank": 1,
+                "credit_limit": 500.0,
+                "risk_sale_order_include": True,
+            }
+        )
```
rrebollo: This is colliding with:

https://github.com/odoo/odoo/blob/9a39b5fc09b2985709781fd20cada45e85e88045/odoo/addons/base/tests/common.py#L32

sale_financial_risk_exclude_order_type/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "Sale Financial Risk Exclude Order Type",
+    "summary": "Exclude configured sale order types from sale risk computation "
+    ""
+    "and blocking",
```
rrebollo: Something strange happened here.

sale_financial_risk_exclude_order_type/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,47 @@
...
+            name="inherit_id"
+            ref="sale_financial_risk.res_config_settings_view_form"
+        />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='allow_overrisk_invoice_validation']/../.."
+                position="after"
+            >
+                <div class="row mt16 o_settings_container">
+                    <div class="o_setting_right_pane">
```
rrebollo: According to a section from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0:

>- Settings views (those inheriting from res.config.settings) have been simplified. Check for the details at https://github.com/odoo/odoo/pull/106425.

>Coming from previous versions, the needed changes are:

>The `<div class="app_settings_block *" ...> element is now <app ...>`.
>The `<div class="o_settings_container *" ...> is now <block ...>`.
>The `<h2>` title is now filled in the title attribute inside the `<block>` tag.
>The `<div class="o_setting_box *" ...>` is now `<setting ...>`.
>The `divs` with classes `o_setting_right_pane` or `o_setting_left_pane` are no more needed and should be removed.
>The fields, labels, paragraphs are put inside `<setting>` tag.

rrebollo (commented):
From a technical perspective, it looks good. However, regarding the design or approach used for the implementation—why not just integrate the risk management directly into `sale_order_type`? I mean, the glue addon is fine, but I don't understand why you went with the company settings approach. Why not simply incorporate the feature into the sale order type itself?

atgalvez08:
> From a technical perspective, it looks good. However, regarding the design or approach used for the implementation—why not just integrate the risk management directly into `sale_order_type`? I mean, the glue addon is fine, but I don't understand why you went with the company settings approach. Why not simply incorporate the feature into the sale order type itself?

Good point, and I agree after revisiting the design.

The initial implementation used a separate configuration model to keep the feature highly configurable and isolated from sale.order.type. The idea was to support company-specific rules without modifying the business model itself.

However, after reviewing the use case more carefully, the additional configuration layer introduced unnecessary complexity for a behavior that is naturally tied to the order type definition itself.

I have now simplified the implementation by moving the configuration directly into sale.order.type through dedicated boolean fields:

- exclude_from_risk
- allow_blocking_bypass

This approach is cleaner because it:

- keeps configuration closer to the business semantics
- removes unnecessary indirection
- reduces maintenance overhead
- simplifies the UX
- avoids an extra model and settings view

The current implementation preserves the same functional behavior while significantly simplifying the architecture.

sale_financial_risk_exclude_order_type/models/res_partner.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        Extend the domain used by sale_financial_risk to compute
+        sale order exposure.
+        Since the base computation runs on sale.order.line, the related
+        field path used here is:
+            order_id.type_id
+        """
+        domain = super()._get_risk_sale_order_domain()
+        sale_type_domain = [("exclude_from_risk", "=", True)]
+        if "company_id" in self.env["sale.order.type"]._fields:
```
carlosdauden: Why is this check carried out?
In what circumstances might `sale.order.type` not contain a `company_id` field?

carlosdauden (changes requested):

sale_financial_risk_exclude_order_type/models/res_partner.py
atgalvez08: Good catch, thanks for the review.

That conditional check came from the previous implementation where the feature was handled through a separate configuration layer and I initially kept the domain construction more defensive.

After simplifying the design and moving the behavior directly into `sale.order.type`, the check became unnecessary because `sale.order.type` already provides `company_id`.

I removed the conditional logic and simplified the domain construction accordingly.

sale_financial_risk_exclude_order_type/models/sale_order_type.py
rrebollo, line 1: Do we really need two flag fields? There seems to be some common ground between the two behaviors. Could you explain this more clearly?

Based on the naming alone, "excluded" appears to result in no sale financial risk behavior, while "allow blocking" seems to do the same. Perhaps the issue is just a lack of clear documentation.

If both paths are legitimate and distinct, I would suggest implementing a single selection field with a range of choices. That said, it's possible that I simply lack enough functional knowledge of the addon.

rrebollo (approved):
Code Review. LGTM!

After reading the documentation, I understand that the two paths address distinct scenarios within the financial risk framework.

PR #568, opened 2026-07-06
[18.0][FIX] account_invoice_overdue_reminder: only take into account posted journal entries for unreconciled warning

alexis-via:
Here is the scenario to reproduce the bug:
1) for customer X, create a customer invoice with a due date in the past and confirm it
2) for customer X, create a customer refund and **cancel** it
3) start overdue wizard

overdue_bug


bguillot (approved):
LGTM

PR #570, opened 2026-07-09
[FIX] account_invoice_overdue_reminder: use correct report action XML ID

jans23:
The method _get_attachment_ids() was calling iaro._render() with 'account.report_invoice_with_payments', which is the XML ID of an ir.ui.view (QWeb template), not an ir.actions.report. Odoo 18 enforces a strict type check in ir.actions.report._get_report(), causing a ValueError at runtime.

Fix: use the correct ir.actions.report XML ID 'account.account_invoices' instead.

### DIFF
--- a/account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
+++ b/account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
@@ -537,7 +537,7 @@ class OverdueReminderStep(models.TransientModel):
         iaro = self.env["ir.actions.report"]
         for inv in self.invoice_ids:
             report_bin, report_format = iaro._render(
-                "account.report_invoice_with_payments", [inv.id]
+                "account.account_invoices", [inv.id]
             )
             filename = f"{inv._get_report_base_filename()}.{report_format}"
             attach = iao.create(


account_invoice_overdue_reminder/wizard/overdue_reminder_wizard.py
```diff
@@ -537,7 +537,7 @@ def _get_attachment_ids(self, mail):
         iaro = self.env["ir.actions.report"]
         for inv in self.invoice_ids:
             report_bin, report_format = iaro._render(
-                "account.report_invoice_with_payments", [inv.id]
+                "account.account_invoices", [inv.id]
```
fkantelberg, line 540: It's not a real issue with the module as the `_render` function uses `_get_report` ([here](https://github.com/OCA/OCB/blob/18.0/odoo/addons/base/models/ir_actions_report.py#L649)) which has a lot of fallbacks.

In a fresh database the following is used:
```xml
<record id="account_invoices" model="ir.actions.report">
    ...
    <field name="report_name">account.report_invoice_with_payments</field>
    <field name="report_file">account.report_invoice_with_payments</field>
    ...
</record>
```

Usually [L664](https://github.com/OCA/OCB/blob/18.0/odoo/addons/base/models/ir_actions_report.py#L664) is used. If you edit the report in your database it goes further and tries to resolve it as a reference and this fails with the given exception.

Nevertheless it change seems to be more inline with the current `account` code as only `account.account_invoices` is referenced directly in multiple cases and not the `report_name`.

fkantelberg (approved):

jans23:
@alexis-via please have a look. Thanks.

PR #574, opened 2026-08-24
[18.0][FIX] account_invoice_overdue_reminder: skip company-dependent conversion when ir_property table is already gone

LorenzoC0:
Skip the company-dependent conversion when the ir_property table no longer exists, scripts from Odoo's official channels (like upgrade.odoo.com or odoo.sh) may already have migrated the table.


LorenzoC0:
Noticed I was using an old `openupgrade` version. Closing.

